Issue #87, opened 2018-05-31
[Suggestion] actor should return a job object after enqueue_job


stegben:
It would be easier to track the job status if the app knows the job id. Also, it would be easier to cancel the job.

samuelcolvin:
Yes, I think would be useful. Same as #53 I think.

samuelcolvin:
fixed by #93

PR #90, opened 2018-07-06
let enqueue_job return job

stegben:
#87

### DIFF
--- a/arq/main.py
+++ b/arq/main.py
@@ -103,7 +103,7 @@ class Actor(RedisMixin, metaclass=ActorMeta):
                 if isinstance(v, CronJob):
                     yield new_v
 
-    async def enqueue_job(self, func_name: str, *args, queue: str=None, **kwargs):
+    async def enqueue_job(self, func_name: str, *args, queue: str=None, **kwargs) -> Job:
         """
         Enqueue a job by pushing the encoded job information into the redis list specified by the queue.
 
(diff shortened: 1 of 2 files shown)


arq/main.py
```diff
@@ -120,18 +120,22 @@ def _bind_decorators(self):
...
             main_logger.debug('%s.%s → %s (called directly)', self.name, func_name, queue)
             data = self.job_class.encode(class_name=self.name, func_name=func_name, args=args, kwargs=kwargs)
             j = self.job_class(data, queue_name=queue)
             await getattr(self, j.func_name).direct(*j.args, **j.kwargs)
+        return j
 
-    def job_future(self, redis, queue: str, func_name: str, *args, **kwargs):
-        return redis.rpush(
+    async def job_future(self, redis, queue: str, func_name: str, *args, **kwargs):
+        data = self.job_class.encode(class_name=self.name, func_name=func_name, args=args, kwargs=kwargs)
```
samuelcolvin: I don't like encoding then decoding here.

Better to modify `Job` to something like

```py
class Job:
    def __init__(self, *, job_id: str=None, queued_at: int=None, ...):
       self.id = self.generate_id(job_id)
       ...

    def encode(self):
        ...
    
    def decode(cls, raw_data: bytes, *, queue_name: str=None, raw_queue: bytes=None): -> Job:
        # like current __init__ method
```

tests/test_main.py
```diff
@@ -203,7 +203,8 @@ class BadActor(Actor):
     class CustomActor(MockRedisDemoActor):
         queues = ['foobar']
     actor = CustomActor(loop=loop)
-    await actor.enqueue_job('add_numbers', 1, 1, queue='foobar')
+    job = await actor.enqueue_job('add_numbers', 1, 1, queue='foobar')
+    assert isinstance(job, Job)
```
samuelcolvin: please add tests to check job id is either random or as set.

arq/main.py
stegben: I add this functionality in #92

samuelcolvin:
replaced by #93

PR #92, opened 2018-07-09
Modify job

stegben:
pre branch of #90

PR #93, opened 2018-07-09
simplifying job encode and decode

samuelcolvin:
replace #90 and #92


samuelcolvin:
We also need methods to allow a job's status to be consulted and to be cancelled, but I guess that's another issue.

stegben:
why don't you merge this and release a new version?

PR #105, opened 2018-11-06
Add missing functions for MockRedis

fokinpv:
I tried to use MockRedis for unit tests and found out that there are no implementation for some functions.


samuelcolvin:
yes, `MockRedis` is incomplete. I'll merge this but I don't expect it to ever be complete.

PR #106, opened 2018-11-12
Use f-string in main_demo code example

Mariatta:
Use f-string because support for 3.5 was dropped and no reason not to.


samuelcolvin:
thanks.

PR #127, opened 2019-05-13
add support for different queues

### LINKED ISSUE #126: Multiple Queues (v.0.16)
How I'm supposed to create multiple queues with arq v0.16?

Looking at old docs I see that you could specify queue name, but now you can't.
What was the purpose of deleting queues from API?

I tried to start workers with different functions, but if worker doesn't see registered function it just deletes job.

So, if I want to use multiple queues on the same redis instance I have to use multiple redis databases (that have only numerical names)?

tsutsarin:
fix #126


arq/worker.py
```diff
@@ -155,6 +157,7 @@ def __init__(
         ctx: Optional[Dict] = None,
     ):
         self.functions: Dict[str, Union[Function, CronJob]] = {f.name: f for f in map(func, functions)}
+        self.queue = queue
```
samuelcolvin: `self.queue_name` please to be consistent with the rest of the code.

arq/worker.py
```diff
@@ -170,6 +175,8 @@ def __init__(
         self.poll_delay_s = to_seconds(poll_delay)
         self.max_tries = max_tries
         self.health_check_interval = to_seconds(health_check_interval)
+        if health_check_key is None:
+            self.health_check_key = self.queue_name + health_check_key_suffix
```
samuelcolvin:
```suggestion
            self.health_check_key = self.queue_name + health_check_key_suffix
        else:
            self.health_check_key = health_check_key
```
:-)

samuelcolvin: also maybe you could add a test for this case.

tsutsarin: added test.

tsutsarin:
ready to merge?

samuelcolvin:
yes, I was just waiting for  tests to pass.

Thank you very much.

samuelcolvin:
released, should be available in pypi in a few minutes.

Thanks again.

tsutsarin:
Thank you too! :tada:

PR #129, opened 2019-05-22
add redis to azure pipelines

samuelcolvin:


samuelcolvin:
redis only seems to work on linux images, not windows. ROFL.

PR #138, opened 2019-08-01
prevent duplicate jobs when result exists

### LINKED ISSUE #137: Resubmission of job with same id causes conflict
Suppose I add a job to a queue and later I add another job with the same job_id then even before the completion of job it returns `JobStatus.complete`. 

example:

```python
import asyncio
from aiohttp import ClientSession
from arq import create_pool

async def download_content(ctx, url):
    session: ClientSession = ctx['session']
    async with session.get(url) as response:
        content = await response.text()
        print(f'{url}: {content:.80}...')
    return len(content)

async def startup(ctx):
    ctx['session'] = ClientSession()

async def shutdown(ctx):
    await ctx['session'].close()


async def main():
    job_list = []
    for url in ('https://facebook.com', 'https://microsoft.com', 'https://github.com'):
        job = await redis.enqueue_job('download_content', url, _job_id=url)
        job_list.append(job)

    for j_ in job_list:
        print( j_.job_id)
        print(await j_.status())
# WorkerSettings defines the settings to use when creating the work,
# it's used by the arq cli
class WorkerSettings:
    functions = [download_content]
    on_startup = startup
    on_shutdown = shutdown

if __name__ == '__main__':
    loop.run_until_complete(main())
```

Running this twice:

```
(sanic) ╭─adam@adams-MacBook-Air ~/sidekik 
╰─$ python arqq_.py
https://facebook.com
JobStatus.queued
https://microsoft.com
JobStatus.queued
https://github.com
JobStatus.queued
(sanic) ╭─adam@adams-MacBook-Air ~/sidekik 
╰─$ python arqq_.py
https://facebook.com
JobStatus.complete
https://microsoft.com
JobStatus.complete
https://github.com
JobStatus.complete
```

samuelcolvin:
fix #137

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,10 @@
 History
 -------
 
+v0.16.1 (unreleased)
+....................
+* prevent duplicate ``job_id`` when job result exists, fix #137
+
 v0.16 (2019-07-30)
 ..................
 * improved error when a job is aborted (eg. function not found)
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -83,8 +83,9 @@ class ArqRedis(Redis):
             pipe.unwatch()
             pipe.watch(job_key)
             job_exists = pipe.exists(job_key)
+            job_result_exists = pipe.exists(result_key_prefix + job_id)
             await pipe.execute()
-            if await job_exists:
+            if await job_exists or await job_result_exists:
                 return
 
             enqueue_time_ms = timestamp_ms()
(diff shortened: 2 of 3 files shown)


powellchristoph:
Greetings! So this change recently broke some code for us and while I can understand the rational for it, how do we work around it? There doesn't appear to be a way to cleanup results once the jobs have been completed?

samuelcolvin:
If you don't want to look at job results set the duration arq keeps results for to a smaller number (or `0` to not keep them at all).

Or you could manually delete the jobs results.

PR #142, opened 2019-08-09
Feat worker limit

### LINKED ISSUE #141: Retrieving new jobs from the queue with a limit
By checking the source code it seems to me that a worker always takes as many jobs as possible from the queue, independently of how many free actors are there:

https://github.com/samuelcolvin/arq/blob/b2d397ae3b07340674f6f0edad144beaf4d4ef54/arq/worker.py#L260-L264

I think this has great potential for improvement. If the queue is large, a worker will keep all the job IDs in memory without being able to execute all of those in a meaningful time. On the contrary, if the semaphore indicates that it's full, why take jobs? (Or maybe we should take just a few.) I was also planning to run multiple instances of my worker, and in that case the parallelism would benefit from each worker taking a subset of jobs that they can manage.

With that in mind I propose that the above call to Redis is changed to:
```python
job_ids = await self.pool.zrangebyscore(self.queue_name, limit=limit)
```

where `limit` is the number of free spots in the semaphore or some multiple of that (ideally configurable, I think the optimal value would mostly depend on the use case but it's probably slightly above `1` in most of them).

I also don't quite understand the usefulness of `now`. I think that in the great majority of the cases, all the enqueued jobs have a timestamp below the current timestamp. But even if they had a slightly higher timestamp, why would a free worker not run it?

rubik:
This PR adds a new parameter to the `Worker` class: `queue_read_limit`. It allows limiting the number of job IDs that are read from the queue at each polling interval.

Redis requires to set both `offset` and `count`, so I've added another instance attribute called `queue_read_offset` that sets the offset to 0 in case `queue_read_limit` is specified, and to `None` otherwise. I haven't added a corresponding parameter for this attribute because I think that in the majority of the cases it's not needed. I think it could be useful to investigate it in case multiple workers are run in parallel, but I haven't tested this so I refrained from complicating the public interface further.

This PR passes all tests except two, which also fail in the `master` branch. So they are most likely unrelated to these changes.

Fixes #141

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -266,20 +266,7 @@ class Worker:
             await self.on_startup(self.ctx)
 
         async for _ in poll(self.poll_delay_s):  # noqa F841
-            async with self.sem:  # don't bother with zrangebyscore until we have "space" to run the jobs
-                now = timestamp_ms()
-                job_ids = await self.pool.zrangebyscore(
-                    self.queue_name, offset=self._queue_read_offset, count=self.queue_read_limit, max=now
-                )
-            await self.run_jobs(job_ids)
-
-            # required to make sure errors in run_job get propagated
-            for t in self.tasks:
-                if t.done():
-                    self.tasks.remove(t)
-                    t.result()
-
-            await self.heart_beat()
+            await self._poll_iteration()
 
             if self.burst:
                 if (
@@ -290,6 +277,28 @@ class Worker:
                 queued_jobs = await self.pool.zcard(self.queue_name)
                 if queued_jobs == 0:
                     return
+            async with self.sem:  # don't bother with zrangebyscore until we have "space" to run the jobs
+                now = timestamp_ms()
+                job_ids = await self.pool.zrangebyscore(
+                    self.queue_name, offset=self._queue_read_offset, count=self.queue_read_limit, max=now
+                )
+            await self.run_jobs(job_ids)
+
+    async def _poll_iteration(self):
+        async with self.sem:  # don't bother with zrangebyscore until we have "space" to run the jobs
+            now = timestamp_ms()
+            job_ids = await self.pool.zrangebyscore(
+                self.queue_name, offset=self._queue_read_offset, count=self.queue_read_limit, max=now
+            )
+        await self.run_jobs(job_ids)
+
+        # required to make sure errors in run_job get propagated
+        for t in self.tasks:
+            if t.done():
+                self.tasks.remove(t)
+                t.result()
+
+        await self.heart_beat()
 
     async def run_jobs(self, job_ids):
         for job_id in job_ids:
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -138,7 +138,8 @@ class Worker:
     :param job_timeout: default job timeout (max run time)
     :param keep_result: default duration to keep job results for
     :param poll_delay: duration between polling the queue for new jobs
-    :param queue_read_limit: the maximum number of jobs to pull from the queue each time it's polled; by default it equals ``max_jobs``
+    :param queue_read_limit: the maximum number of jobs to pull from the queue each time it's polled; by default it
+                             equals ``max_jobs``
     :param max_tries: default maximum number of times to retry a job
     :param health_check_interval: how often to set the health check key
     :param health_check_key: redis key under which health check is set
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -138,7 +138,7 @@ class Worker:
     :param job_timeout: default job timeout (max run time)
     :param keep_result: default duration to keep job results for
     :param poll_delay: duration between polling the queue for new jobs
-    :param queue_read_limit: the maximum number of jobs to pull from the queue each time it's polled
+    :param queue_read_limit: the maximum number of jobs to pull from the queue each time it's polled; by default it equals ``max_jobs``
     :param max_tries: default maximum number of times to retry a job
     :param health_check_interval: how often to set the health check key
     :param health_check_key: redis key under which health check is set
@@ -186,10 +186,8 @@ class Worker:
         self.poll_delay_s = to_seconds(poll_delay)
         self.queue_read_limit = queue_read_limit
         if self.queue_read_limit is None:
-            # Redis requires offset and count to be both set or both unset
-            self.queue_read_offset = None
-        else:
-            self.queue_read_offset = 0
+            self.queue_read_limit = max_jobs
+        self._queue_read_offset = 0
         self.max_tries = max_tries
         self.health_check_interval = to_seconds(health_check_interval)
         if health_check_key is None:
@@ -271,7 +269,7 @@ class Worker:
             async with self.sem:  # don't bother with zrangebyscore until we have "space" to run the jobs
                 now = timestamp_ms()
                 job_ids = await self.pool.zrangebyscore(
-                    self.queue_name, offset=self.queue_read_offset, count=self.queue_read_limit, max=now
+                    self.queue_name, offset=self._queue_read_offset, count=self.queue_read_limit, max=now
                 )
             await self.run_jobs(job_ids)
 
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -185,9 +185,7 @@ class Worker:
         self.job_timeout_s = to_seconds(job_timeout)
         self.keep_result_s = to_seconds(keep_result)
         self.poll_delay_s = to_seconds(poll_delay)
-        self.queue_read_limit = queue_read_limit
-        if self.queue_read_limit is None:
-            self.queue_read_limit = max_jobs
+        self.queue_read_limit = queue_read_limit or max_jobs
         self._queue_read_offset = 0
         self.max_tries = max_tries
         self.health_check_interval = to_seconds(health_check_interval)
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -185,6 +185,11 @@ class Worker:
         self.keep_result_s = to_seconds(keep_result)
         self.poll_delay_s = to_seconds(poll_delay)
         self.queue_read_limit = queue_read_limit
+        if self.queue_read_limit is None:
+            # Redis requires offset and count to be both set or both unset
+            self.queue_read_offset = None
+        else:
+            self.queue_read_offset = 0
         self.max_tries = max_tries
         self.health_check_interval = to_seconds(health_check_interval)
         if health_check_key is None:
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -18,9 +18,9 @@
 async def worker(arq_redis):
     worker_: Worker = None
 
-    def create(functions=[], burst=True, poll_delay=0, **kwargs):
+    def create(functions=[], burst=True, poll_delay=0, queue_read_limit=40, **kwargs):
         nonlocal worker_
-        worker_ = Worker(functions=functions, redis_pool=arq_redis, burst=burst, poll_delay=poll_delay, **kwargs)
+        worker_ = Worker(functions=functions, redis_pool=arq_redis, burst=burst, poll_delay=poll_delay, queue_read_limit=queue_read_limit, **kwargs)
```
samuelcolvin: The only thing causing tests to fail on CI is this line which is too long, please run `make format`.

rubik: @samuelcolvin Thanks, done. On my system two tests, `test_no_jobs` and `test_health_check_direct` always fail. I don't know the reason, but this is the error in both cases:

```
self = <asyncio.unix_events._UnixDefaultEventLoopPolicy object at 0x7fc091335e10>

    def get_event_loop(self):
        """Get the event loop.
    
        This may be None or an instance of EventLoop.
        """
        if (self._local._loop is None and
                not self._local._set_called and
                isinstance(threading.current_thread(), threading._MainThread)):
            self.set_event_loop(self.new_event_loop())
    
        if self._local._loop is None:
            raise RuntimeError('There is no current event loop in thread %r.'
>                              % threading.current_thread().name)
E           RuntimeError: There is no current event loop in thread 'MainThread'.

/usr/lib64/python3.7/asyncio/events.py:644: RuntimeError
```

samuelcolvin: humm, not sure, seems to work fine for me:

```
(env) arq 0  11ms ➤  pytest -k test_health_check_direct
Test session starts (platform: linux, Python 3.7.4, pytest 4.4.1, pytest-sugar 0.9.2)
rootdir: /home/samuel/code/arq, inifile: setup.cfg, testpaths: tests
plugins: mock-1.10.4, toolbox-0.4, cov-2.6.1, timeout-1.3.3, sugar-0.9.2, aiohttp-0.3.0
timeout: 5.0s
timeout method: signal
timeout func_only: False
collecting ... 
 tests/test_worker.py ✓                                          100% ██████████

Results (0.06s):
       1 passed
      97 deselected
(env) arq 0 419ms ➤  
```

Do you have the same plugins and versions?

rubik: @samuelcolvin I have no idea why it does not work. I have the exact same version of everything, even Python. It should be tracked in a new issue, because I get the same on `master`.

```
 pytest -k test_health_check_direct
Test session starts (platform: linux, Python 3.7.4, pytest 4.4.1, pytest-sugar 0.9.2)
rootdir: /home/miki/exp/arq, inifile: setup.cfg, testpaths: tests
plugins: mock-1.10.4, sugar-0.9.2, timeout-1.3.3, cov-2.6.1, aiohttp-0.3.0, toolbox-0.4
timeout: 5.0s
timeout method: signal
timeout func_only: False
collecting ... 

―――――――――――――――――――――――――――――――――――― test_health_check_direct ――――――――――――――――――――――――――――――――――――

loop = <_UnixSelectorEventLoop running=False closed=False debug=False>

    def test_health_check_direct(loop):
        class Settings:
            pass
    
>       assert check_health(Settings) == 1

tests/test_worker.py:41: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
arq/worker.py:610: in check_health
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <asyncio.unix_events._UnixDefaultEventLoopPolicy object at 0x7f080bc70210>

    def get_event_loop(self):
        """Get the event loop.
    
        This may be None or an instance of EventLoop.
        """
        if (self._local._loop is None and
                not self._local._set_called and
                isinstance(threading.current_thread(), threading._MainThread)):
            self.set_event_loop(self.new_event_loop())
    
        if self._local._loop is None:
            raise RuntimeError('There is no current event loop in thread %r.'
>                              % threading.current_thread().name)
E           RuntimeError: There is no current event loop in thread 'MainThread'.

/usr/lib64/python3.7/asyncio/events.py:644: RuntimeError

 tests/test_worker.py ⨯                                                            100% ██████████

Results (0.31s):
       1 failed
         - tests/test_worker.py:37 test_health_check_direct
      97 deselected
```

samuelcolvin: new issue please.

This won't be related to pytest-asyncio, since we use pytest-aiohttp for async tests.

I also doubt it's related to the aiohttp issue since that's quite old, the event loop logic changed in `3.5.3` and most libraries have gradually changed how they work with `loop` since they've ceased to support `<=3.5.2`.

rubik: I have opened issue #144. Note that in the pytest-asyncio issue pytest-aiohttp is mentioned.

arq/worker.py
```diff
@@ -158,6 +161,7 @@ def __init__(
         job_timeout: SecondsTimedelta = 300,
         keep_result: SecondsTimedelta = 3600,
         poll_delay: SecondsTimedelta = 0.5,
+        queue_read_limit: Optional[int] = None,
```
samuelcolvin: maybe this shouldn't be optional?

Is there any reason not to set it to `1000` or `max_jobs`?

rubik:
@samuelcolvin Yes, tests are definitely needed. I added them in the second-to-last commit. I also added the history note.

arq/worker.py
samuelcolvin: @rubik did you see this comment? What do you think?

arq/worker.py
```diff
@@ -280,6 +278,28 @@ def run(self) -> None:
                 queued_jobs = await self.pool.zcard(self.queue_name)
                 if queued_jobs == 0:
                     return
+            async with self.sem:  # don't bother with zrangebyscore until we have "space" to run the jobs
```
samuelcolvin: I don't understand why we're now doing `zrangebyscore` in two different places.

Is this really required?

tests/test_worker.py
```diff
@@ -449,3 +450,55 @@ class Foo:
...
     assert await arq_redis.enqueue_job('foobar', _job_id='job_id') is None
+
+
+async def test_queue_read_limit_equals_max_jobs(arq_redis: ArqRedis, worker):
+    for _ in range(4):
+        await arq_redis.enqueue_job('foobar')
+
+    assert await arq_redis.zcard(default_queue_name) == 4
+    worker: Worker = worker(functions=[foobar], max_jobs=2)
+    worker.pool = await create_pool(worker.redis_settings)
```
samuelcolvin: why do you need this?

tests/test_worker.py
```diff
@@ -449,3 +450,55 @@ class Foo:
...
+    assert worker.jobs_complete == 0
+    assert worker.jobs_failed == 0
+    assert worker.jobs_retried == 0
+
+    await worker._poll_iteration()
+    await asyncio.sleep(0.01)
```
samuelcolvin: what is this sleep doing? I doubt we need it.

arq/worker.py
rubik: @samuelcolvin I agree, yes. One of the last commits sets it to `max_jobs` if it's None.

tests/test_worker.py
rubik: @samuelcolvin It introduces a small delay because if Redis and Python are not fast enough, the assert below is not verified. In fact, if I remove it the test fails on my machine.

arq/worker.py
samuelcolvin: okay, so the type shouldn't be `Optional`, on the class

should be:
```py
self.queue_read_limit: int = queue_read_limit or max_jobs
```

tests/test_worker.py
samuelcolvin: okay, I would guess you might need a big delay to avoid intermittent failures on CI where resources are more constrained (I've had problems like this a lot before), would increase to `0.1`.

tests/test_worker.py
rubik: @samuelcolvin My understanding was that the pool is initialized in `Worker.main`, which we don't call in this test. It seems that it's initialized separately too, so I removed this line.

arq/worker.py
rubik: @samuelcolvin This was introduced by mistake, it's absolutely not required. The only thing needed is what is inside `_poll_iteration`. This has been fixed now.

tests/test_worker.py
samuelcolvin: I think the point is that the `worker` fixture uses the standard redis connection, thus this isn't required.

arq/worker.py
rubik: @samuelcolvin I don't follow. In your line with `or` you allow `queue_read_limit` to be `None`. That means that we need to keep `Optional` in the type hint. If the user does not specify it, it's `None`. I changed the assignment to be a one-liner like yours.

tests/test_worker.py
rubik: @samuelcolvin It makes sense. I increased it to `0.1`.

arq/worker.py
samuelcolvin: What you've done is correct.

My point is that if you do:

```py
foo: Optional[int] = 1

bar = foo

if bar is None:
    bar = 123
```
Mypy reads this as implicitly setting the type of `bar` to `Optional[int]`, that doesn't change in the `if` block.

If instead you do

```py
foo: Optional[int] = 1
bar = foo or 123
```
The type of `bar` is `int`, this is what we do above, except we're explicit and add a type hint.

rubik: @samuelcolvin Now I understand, thanks for the explanation.

rubik:
@samuelcolvin I have fixed the linting errors so that the checks can pass. The errors on Python 3.8 remain, but they are not related to this PR. I am not able to understand what causes them.

samuelcolvin:
no problem, looks like aioredis is not compatible with python 3.8.

tests/test_worker.py
samuelcolvin: looks like you've only updated this number in some places.

PR #143, opened 2019-08-10
Custom serializers

rubik:
## Changes

I would like to propose this change, which allows the user to specify custom serializers in alternative to pickle, which is left as the default.

The public API is changed only very slightly: there are new `_serialize` and `_deserialize` arguments to `create_pool` and `Worker`. The rest of the changes essentially reduces to passing these to the pickle functions.

For backward compatibility, I have left the `pickle_jobs`, etc. functions in `arq.jobs` with the same name, even though the name is slightly inappropriate now.

## Rationale

The main reason I am proposing this change is to allow better memory usage. The default serialization method, pickle, is quite space inefficient. Since the job data is essentially JSON, we can attain much better memory usage by switching to serialization methods that are more appropriate.

In my case, by using [MsgPack](https://msgpack.org/) instead of Pickle, I am seeing an improvement of 47%-49% in memory usage which is quite significant. At tens to hundreds of thousands of tasks this means hundreds of MBs if not GBs in savings.

## Usage

The usage is quite simple. Existing code works as usual, but one can use a different serialization method as follows:

```python
import msgpack
from arq.connections import create_pool, RedisSettings

pool = await create_pool(RedisSettings(), _serialize=msgpack.packb, _deserialize=msgpack.unpackb)

WorkerSettings:
    serialize = msgpack.packb
    deserialize = msgpack.unpackb
```

## Tests and documentation
The tests pass, except two tests that also fail on master. If this change is approved, I can write a paragraph in the documentation on custom serialization functions.

### DIFF
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -3,7 +3,7 @@ import logging
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from operator import attrgetter
-from typing import Any, List, Optional, Union
+from typing import Any, List, Optional, Union, Callable
 from uuid import uuid4
 
 import aioredis
@@ -43,7 +43,23 @@ expires_extra_ms = 86_400_000
 class ArqRedis(Redis):
     """
     Thin subclass of ``aioredis.Redis`` which adds :func:`arq.connections.enqueue_job`.
+
+    :param redis_settings: an instance of ``arq.connections.RedisSettings``.
+    :param _serialize: a function that serializes Python objects to bytes, defaults to pickle.dumps
+    :param _deserialize: a function that deserializes bytes into Python objects, defaults to pickle.loads
+    :param kwargs: keyword arguments directly passed to ``aioredis.Redis``.
     """
+
+    def __init__(
+        self,
+        redis_settings,
+        _serialize: Optional[Callable[[Any], bytes]] = None,
+        _deserialize: Optional[Callable[[bytes], Any]] = None,
+        **kwargs,
+    ) -> None:
+        self._serialize = _serialize
+        self._deserialize = _deserialize
+        super().__init__(redis_settings, **kwargs)
 
     async def enqueue_job(
         self,
@@ -98,7 +114,7 @@ class ArqRedis(Redis):
 
             expires_ms = expires_ms or score - enqueue_time_ms + expires_extra_ms
 
-            job = pickle_job(function, args, kwargs, _job_try, enqueue_time_ms)
+            job = pickle_job(function, args, kwargs, _job_try, enqueue_time_ms, _serialize=self._serialize)
             tr = conn.multi_exec()
             tr.psetex(job_key, expires_ms, job)
             tr.zadd(_queue_name, score, job_id)
@@ -107,11 +123,11 @@ class ArqRedis(Redis):
             except MultiExecError:
                 # job got enqueued since we checked 'job_exists'
                 return
-        return Job(job_id, self)
+        return Job(job_id, redis=self, _deserialize=self._deserialize)
 
     async def _get_job_result(self, key):
         job_id = key[len(result_key_prefix) :]
-        job = Job(job_id, self)
+        job = Job(job_id, self, _deserialize=self._deserialize)
         r = await job.result_info()
         r.job_id = job_id
         return r
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -54,12 +54,12 @@ class ArqRedis(Redis):
     def __init__(
         self,
         pool_or_conn,
-        _job_serializer: Optional[Serializer] = None,
-        _job_deserializer: Optional[Deserializer] = None,
+        job_serializer: Optional[Serializer] = None,
+        job_deserializer: Optional[Deserializer] = None,
         **kwargs,
     ) -> None:
-        self._job_serializer = _job_serializer
-        self._job_deserializer = _job_deserializer
+        self.job_serializer = job_serializer
+        self.job_deserializer = job_deserializer
         super().__init__(pool_or_conn, **kwargs)
 
     async def enqueue_job(
@@ -115,7 +115,7 @@ class ArqRedis(Redis):
 
             expires_ms = expires_ms or score - enqueue_time_ms + expires_extra_ms
 
-            job = serialize_job(function, args, kwargs, _job_try, enqueue_time_ms, serializer=self._job_serializer)
+            job = serialize_job(function, args, kwargs, _job_try, enqueue_time_ms, serializer=self.job_serializer)
             tr = conn.multi_exec()
             tr.psetex(job_key, expires_ms, job)
             tr.zadd(_queue_name, score, job_id)
@@ -124,11 +124,11 @@ class ArqRedis(Redis):
             except MultiExecError:
                 # job got enqueued since we checked 'job_exists'
                 return
-        return Job(job_id, redis=self, _deserializer=self._job_deserializer)
+        return Job(job_id, redis=self, _deserializer=self.job_deserializer)
 
     async def _get_job_result(self, key):
         job_id = key[len(result_key_prefix) :]
-        job = Job(job_id, self, _deserializer=self._job_deserializer)
+        job = Job(job_id, self, _deserializer=self.job_deserializer)
         r = await job.result_info()
         r.job_id = job_id
         return r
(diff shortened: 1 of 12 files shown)


arq/connections.py
```diff
@@ -43,8 +44,24 @@ def __repr__(self):
...
+    :param redis_settings: an instance of ``arq.connections.RedisSettings``.
+    :param _serialize: a function that serializes Python objects to bytes, defaults to pickle.dumps
+    :param _deserialize: a function that deserializes bytes into Python objects, defaults to pickle.loads
+    :param kwargs: keyword arguments directly passed to ``aioredis.Redis``.
     """
 
+    def __init__(
+        self,
+        pool_or_conn,
+        _serialize: Optional[Callable[[Any], bytes]] = None,
```
samuelcolvin:
```suggestion
        _job_serializer: Optional[Callable[[Any], bytes]] = None,
```
Might be even clearer?

samuelcolvin: and same obviously for `_deserialize` -> `_job_deserializer`.

arq/jobs.py
```diff
@@ -119,16 +126,30 @@ def __repr__(self):
...
+class SerializationError(RuntimeError):
     pass
 
 
-def pickle_job(function_name: str, args: tuple, kwargs: dict, job_try: int, enqueue_time_ms: int):
+class PickleError(SerializationError):
+    pass
+
+
+def pickle_job(
```
samuelcolvin: let's rename `unpickle_job` to `deserialize_job`, better to get it right now.

Change should be pretty explicit and easy to fix.

arq/jobs.py
```diff
@@ -119,16 +126,30 @@ def __repr__(self):
         return f'<arq job {self.job_id}>'
 
 
-class PickleError(RuntimeError):
+class SerializationError(RuntimeError):
     pass
 
 
-def pickle_job(function_name: str, args: tuple, kwargs: dict, job_try: int, enqueue_time_ms: int):
+class PickleError(SerializationError):
```
samuelcolvin: remove.

arq/connections.py
```diff
@@ -43,8 +44,24 @@ def __repr__(self):
...
+    :param _job_serializer: a function that serializes Python objects to bytes, defaults to pickle.dumps
+    :param _job_deserializer: a function that deserializes bytes into Python objects, defaults to pickle.loads
+    :param kwargs: keyword arguments directly passed to ``aioredis.Redis``.
     """
 
+    def __init__(
+        self,
+        pool_or_conn,
+        _job_serializer: Optional[Callable[[Any], bytes]] = None,
+        _job_deserializer: Optional[Callable[[bytes], Any]] = None,
```
samuelcolvin:
```suggestion
        _job_deserializer: Optional[Callable[[bytes], Dict[str, Any]] = None,
```
Or am I missing something?

arq/connections.py
```diff
@@ -43,8 +44,24 @@ def __repr__(self):
...
+    :param redis_settings: an instance of ``arq.connections.RedisSettings``.
+    :param _job_serializer: a function that serializes Python objects to bytes, defaults to pickle.dumps
+    :param _job_deserializer: a function that deserializes bytes into Python objects, defaults to pickle.loads
+    :param kwargs: keyword arguments directly passed to ``aioredis.Redis``.
     """
 
+    def __init__(
+        self,
+        pool_or_conn,
+        _job_serializer: Optional[Callable[[Any], bytes]] = None,
```
samuelcolvin:
```suggestion
        _job_serializer: Optional[Callable[[Dict[str, Any]], bytes]] = None,
```
I think?

samuelcolvin: and obviously elsewhere where it's used, might be worth having these types defined once, eg. in `jobs.py`?

arq/jobs.py
```diff
@@ -119,19 +126,29 @@ def __repr__(self):
...
 
-def pickle_job(function_name: str, args: tuple, kwargs: dict, job_try: int, enqueue_time_ms: int):
+def serialize_job(
+    function_name: str,
+    args: tuple,
+    kwargs: dict,
+    job_try: int,
+    enqueue_time_ms: int,
+    *,
+    _serialize: Optional[Callable[[Any], bytes]] = None,
```
samuelcolvin:
```suggestion
    serialize: Optional[Callable[[Any], bytes]] = None,
```
since this is explicitly required as a keyword arg, I don't think you need to make it "private".

tests/test_worker.py
```diff
@@ -449,3 +451,26 @@ class Foo:
...
+    await worker.main()
+    assert await j.result() == 42
+    r = await j.info()
+    assert r.result == 42
+
+
+async def test_incompatible_serializers_1(arq_redis_msgpack: ArqRedis, worker):
+    await arq_redis_msgpack.enqueue_job('foobar', _job_id='job_id')
+    worker: Worker = worker(functions=[foobar])
+    with pytest.raises(SerializationError) as exc_info:
```
samuelcolvin: what's the error here? Might be good to add `assert str(exc_info.value) == ...` just to document what the error is.

And below.

rubik:
@samuelcolvin I addressed your suggestions. You said that `serializer` should be public and therefore starting without underscore for the functions in `arq.jobs`. Following that reasoning, the same correction needs to be applied to `arq.connections.create_pool`. That also makes the arguments equal to the parameters in `arq.worker.Worker`. That was done in a separate commit, https://github.com/samuelcolvin/arq/pull/143/commits/0cdaf4bbb0893b5d8c769d7e248cb4e9a4ee2f4a, so if you don't agree I can revert it.

I don't think the HISTORY.rst change belongs to this PR and you should do it in a separate commit, also because I don't know when you will release the next version (even though I hope very soon).

samuelcolvin:
The way I work every commit to code has an entry in `HISTORY.rst`, it's the case for all projects so I'm afraid it's a requirement.

Please add the following to history

```
v0.17 (unreleased)
..................
* custom serializers, eg. to use msgpack rather than pickle, #143 by @rubik 

```

tests/test_worker.py
```diff
@@ -449,3 +451,32 @@ class Foo:
...
+        await worker.main()
+        assert exc_info.value.startswith('unable to deserialize job: ')
```
samuelcolvin:
```suggestion
    assert exc_info.value.startswith('unable to deserialize job: ')
```
this is indented wrongly I think.

tests/test_worker.py
```diff
@@ -449,3 +451,32 @@ class Foo:
...
+
+
+async def test_incompatible_serializers_2(arq_redis: ArqRedis, worker):
+    await arq_redis.enqueue_job('foobar', _job_id='job_id')
+    worker: Worker = worker(
+        functions=[foobar], job_serializer=msgpack.packb, job_deserializer=functools.partial(msgpack.unpackb, raw=False)
+    )
+    with pytest.raises(SerializationError) as exc_info:
+        await worker.main()
+        assert exc_info.value.startswith('unable to deserialize job: ')
```
samuelcolvin:
```suggestion
    assert exc_info.value.startswith('unable to deserialize job: ')
```

rubik:
@samuelcolvin Done. Then after you merge this PR I'll add it for the other one.

rubik:
@samuelcolvin I fixed the code as per your latest feedback.

rubik:
It seems some tests fail on Python 3.8, but I don't think the errors are related to this PR.

arq/connections.py
```diff
@@ -43,8 +44,24 @@ def __repr__(self):
 class ArqRedis(Redis):
     """
     Thin subclass of ``aioredis.Redis`` which adds :func:`arq.connections.enqueue_job`.
+
+    :param redis_settings: an instance of ``arq.connections.RedisSettings``.
+    :param _job_serializer: a function that serializes Python objects to bytes, defaults to pickle.dumps
```
samuelcolvin:
```suggestion
    :param job_serializer: a function that serializes Python objects to bytes, defaults to pickle.dumps
```

arq/connections.py
```diff
@@ -43,8 +44,24 @@ def __repr__(self):
...
+    :param _job_deserializer: a function that deserializes bytes into Python objects, defaults to pickle.loads
```
samuelcolvin:
```suggestion
    :param job_deserializer: a function that deserializes bytes into Python objects, defaults to pickle.loads
```

rubik: @samuelcolvin My bad, it's fixed.

samuelcolvin:
Great, thanks. I'll deploy as soon as the other PR is passing and I can merge it.

rubik:
@samuelcolvin Thanks for taking the time to review it multiple times.

PR #149, opened 2019-08-21
better logic for deserializing

samuelcolvin:
@rubik, some changes to deserialization. You might want to check this doesn't conflict with your usage of msgpack, it shouldn't but better to check.


rubik:
@samuelcolvin All good. In fact, I wanted to propose these changes myself but then I didn't find the time to do it.

PR #151, opened 2019-08-21
Support for Redis Sentinel

pratheekrebala:
Hey Sam,

Apologies about the delay on this. I think this should do the trick to get the script to work with Sentinel.

Sidenote - is the `commands_factory` doing anything beyond calling `ArqRedis(pool)`? The `commands_factory` argument is unavailable for `create_sentinel_pool` and is [marked as deprecated](https://aioredis.readthedocs.io/en/v1.2.0/api_reference.html?highlight=commands_factory#connections-pool) in some places. I'm unsure if this argument is deprecated universally or just in that specific context.

I'm calling `ArqRedis(pool)` instead to generalize the initialization of the pool. Hope this is ok.

Thanks,
Pratheek


arq/connections.py
```diff
@@ -169,18 +170,31 @@ def __init__(
...
     """
     settings = settings or RedisSettings()
-    addr = settings.host, settings.port
+
     try:
-        pool = await aioredis.create_redis_pool(
+        if settings.sentinels:
+            async def pool_factory(*args, **kwargs):
+                client = await aioredis.sentinel.create_sentinel_pool(*args, **kwargs)
+                return client.master_for("mymaster")
```
samuelcolvin: single quotes please.

samuelcolvin: what is `mymaster` will this always be the value used?

samuelcolvin: then I guess it needs to be configurable via `RedisSettings`.

arq/connections.py
```diff
@@ -32,6 +33,10 @@ class RedisSettings:
     conn_timeout: int = 1
     conn_retries: int = 5
     conn_retry_delay: int = 1
+    sentinel: dict = {
```
samuelcolvin: I think better to have these as two different fields.

arq/connections.py
```diff
@@ -169,18 +174,31 @@ def __init__(
     thus allowing job enqueuing.
     """
     settings = settings or RedisSettings()
-    addr = settings.host, settings.port
+
     try:
-        pool = await aioredis.create_redis_pool(
+        if settings.sentinel.master:
```
samuelcolvin: I think this wrong.

arq/connections.py
```diff
@@ -169,18 +174,31 @@ def __init__(
...
+                return client.master_for(settings.sentinel['master'])
+
+            addr = [(s['host'], s['port']) for s in settings.sentinel['hosts']]
+
+            settings.host = addr[0][0]
+            settings.port = addr[0][1]
+        
+        else:
+            pool_factory = aioredis.create_redis_pool
+            addr = settings.host, settings.port
```
samuelcolvin: I think all the above logic can be moved out of the try block?

samuelcolvin:
> However, travis doesn't have a easy way to test against a Redis HA setup.
> 
> Any ideas?

what about [this](https://github.com/Grokzen/travis-redis-cluster)?

Alternatively just add a test which calls `create_pool` with a master and multiple hosts, and expects the call to fail as you'd expect.

arq/connections.py
```diff
@@ -169,24 +171,34 @@ def __init__(
...
+
+    if settings.sentinel:
+        addr = settings.host
+
+        async def pool_factory(*args, **kwargs):
+            client = await aioredis.sentinel.create_sentinel_pool(*args, **kwargs)
+            return client.master_for(settings.sentinel_master)
+
+    else:
+        pool_factory = aioredis.create_redis_pool
```
samuelcolvin: I think this should be `aioredis.pool.create_pool`, that might be why the test is failing.

pratheekrebala:
Hi @samuelcolvin,

sorry I dropped the ball on this. I think this should be good to go. could you let me know if everything looks OK to merge? Happy to work more on this either this evening or tomorrow.

Thanks!

samuelcolvin:
great, thank you.

PR #154, opened 2019-09-05
task processing speedup

solarw:
use `poll_delay` in case of no task in queues, otherway consume tasks without delay

see no reason to sleep after every bunch of tasks fetched from redis.
would be nice to sleep if no task to process left in redis.


samuelcolvin:
Sorry for the slow response.

I think this is not necessary, `pool` will already limit the rate at which the sorted set is pooled. If we've delayed by more than `self.poll_delay_s` it's correct to immediately check for more jobs.

PR #157, opened 2019-10-30
Fix cron scheduling.

Tinche:
If your worker settings specify a non-default queue, cron jobs won't run.


samuelcolvin:
you're missing spaces, that will be wrong.

samuelcolvin:
superceded by #159

PR #163, opened 2019-11-15
Add py.typed file

steinitzu:
So mypy doesn't complain about missing imports :)


samuelcolvin:
Thanks a lot, could you add an entry (might require a new section) to `HISTORY.rst`

steinitzu:
Sure, I added a new patch version. Hope that's ok

PR #170, opened 2020-02-11
Fix workers concurrency

CorentinCAM:
When running several workers, the following situation can occur:

Let's say two workers are running and each one allow two jobs maximum at the same time.

- The first worker takes the two firsts jobs of the queue
- The second worker get the two firsts jobs, but sees that they are ran by another worker. It skips both of them and start again with the same jobs until they are completed by the first worker.
- The second worker will keep doing that until it is the first one to see a job.

To fix this problem, we just increment the `_queue_read_offset` variable, which will allow the worker to get jobs after the ones that are currently running elsewhere.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,10 @@
 History
 -------
 
+v0.18.5 (2020-02-11)
+....................
+* Fix workers concurrency
+
 v0.18.4 (2019-12-19)
 ....................
 * Add ``py.typed`` file to tell mypy the package has type hints, #163
--- a/arq/version.py
+++ b/arq/version.py
@@ -2,4 +2,4 @@ from distutils.version import StrictVersion
 
 __all__ = ['VERSION']
 
-VERSION = StrictVersion('0.18.4')
+VERSION = StrictVersion('0.18.5')
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -295,6 +295,8 @@ class Worker:
             job_ids = await self.pool.zrangebyscore(
                 self.queue_name, offset=self._queue_read_offset, count=count, max=now
             )
+            if job_ids == [] and self._queue_read_offset > 0:
+                self._queue_read_offset = 0
         await self.run_jobs(job_ids)
 
         for t in self.tasks:
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -319,8 +319,10 @@ class Worker:
                 if ongoing_exists or not score:
                     # job already started elsewhere, or already finished and removed from queue
                     self.sem.release()
+                    self._queue_read_offset += 1
                     continue
 
+                self._queue_read_offset = 0
                 tr = conn.multi_exec()
                 tr.setex(in_progress_key, self.in_progress_timeout_s, b'1')
                 try:


samuelcolvin:
I see a problem, but I think you have it defined slightly wrong.

Every worker will take **all** pending job ids and run them unless they're already running elsewhere.

Thus there could be a problem where two jobs both happen on the same worker instead of being split between multiple workers, but not exactly the blocking scenario you describe.

If this solution does solve the problem:
1. There needs to be a better explanation of how it will avoid the problem
2. There needs to be a test demonstrating the fix, though I can imagine this would be very difficult to set up

CorentinCAM:
I hope I understood how your code works, otherwise please correct me.

I think `job_ids` are selected in `_poll_iteration` method (with `pool.zrangebyscore`), according to `queue_name`, `_queue_read_offset` and `count`, which is basically the `max_jobs` parameter. So only the `max_jobs` first jobs will be selected. Those jobs won't be removed from the pending jobs queue until they are completed by a worker.

So when a worker takes the first n jobs, the next worker to poll the queue will get the m first ones, and if m is lower than n all the jobs it will get will be taken by the first worker.

My solution is to increase the `_queue_read_offset` when a worker gets a job that is already taken, so next time it will poll the queue, it will get the next jobs in the queue (until count + 1).

I'll try to give you an example:

First time:
- First worker polls the queue and takes jobs 0 and 1
- Second worker polls the queue and get jobs 0 and 1, both already taken, it increases `_queue_read_offset` by 2.

Next iteration:
- First worker is still working on jobs 0 and 1
- Second worker polls the queue **from 2** and gets jobs 2 and 3, that are free
- As it gets at least one free job, it resets `_queue_read_offset` to 0 so it will poll the first jobs next time

An other solution could be to increase `queue_read_limit` to get more jobs after a fail but I am not sure about how to implement it and I am pretty sure this solution is more simple.

CorentinCAM:
I realised an issue was present with my code.

When a worker increased its offset so much that it was getting no jobs anymore, and new jobs were added below its offset, it was never taking them. Now when no jobs are found and offset is not zero, it is reset to zero in order to find potential new jobs.

samuelcolvin:
Humm, I'm still not convinced this is the solution and either way we need tests.

Can you explain how to reproduce this error and I'll try to fix it.

CorentinCAM:
Ok, here is a demo:

demo.py:
```
import time
import asyncio
from arq import create_pool

async def wait(ctx, sec):
    time.sleep(sec)

async def main():

    await redis.enqueue_job('wait', 2, _job_id='job0')
    await redis.enqueue_job('wait', 2, _job_id='job1')
    await redis.enqueue_job('wait', 2, _job_id='job2')
    await redis.enqueue_job('wait', 2, _job_id='job3')

class WorkerSettings:
    functions = [wait]
    max_jobs = 1

if __name__ == '__main__':
    loop.run_until_complete(main())
```

Now start two workers with `arq demo.WorkerSettings` and then start `python demo.py`

The first worker will take all the jobs, as the seen in the following logs:

- From worker 1:
```
11:30:48: Starting worker for 1 functions: wait
11:30:48: redis_version=5.0.7 mem_usage=834.68K clients_connected=1 db_keys=0
11:32:19:   0.03s → job0:wait(2)
11:32:21:   2.00s ← job0:wait ● 
11:32:21:   2.04s → job1:wait(2) delayed=2.04s
11:32:23:   2.00s ← job1:wait ● 
11:32:23:   4.06s → job2:wait(2) delayed=4.06s
11:32:25:   2.00s ← job2:wait ● 
11:32:25:   6.07s → job3:wait(2) delayed=6.07s
11:32:27:   2.00s ← job3:wait ●
```
- From worker 2:
```
11:30:52: Starting worker for 1 functions: wait
11:30:52: redis_version=5.0.7 mem_usage=855.37K clients_connected=2 db_keys=1
```

Without max_jobs parameter everything is fine but when you want to limit it it doesn't work at all.

Here are the logs for the same execution with my solution:
- From worker 1:
```
11:35:52: Starting worker for 1 functions: wait
11:35:52: redis_version=5.0.7 mem_usage=834.86K clients_connected=1 db_keys=0
11:35:57:   0.28s → job0:wait(2)
11:35:59:   2.00s ← job0:wait ● 
11:36:00:   2.79s → job2:wait(2) delayed=2.79s
11:36:02:   2.00s ← job2:wait ●
```
- From worker 2:
```
11:35:54: Starting worker for 1 functions: wait
11:35:54: redis_version=5.0.7 mem_usage=855.55K clients_connected=2 db_keys=1
11:35:58:   0.95s → job1:wait(2)
11:36:00:   2.00s ← job1:wait ● 
11:36:00:   3.46s → job3:wait(2) delayed=3.46s
11:36:02:   2.00s ← job3:wait ●
```

samuelcolvin:
Thank so much for working on this and demonstrating that there's a problem.

I worked through it and found a much simpler fix, see #180.

If you have any questions about my solution, please ask on #180.

Issue #172, opened 2020-02-21
How to cancel a defer job?


long2ice:
If a job enqueue defer already, then how to cancel it ?

samuelcolvin:
Thank so much for working on this. Sorry I never replied to the issue.

Overall the PR looks good, just a few more things to do.

PR #173, opened 2020-03-24
possibility to abort long running jobs

david-e:
This patch permits to abort deferred jobs.
It will fix issue #172

### DIFF
--- a/arq/jobs.py
+++ b/arq/jobs.py
@@ -6,8 +6,7 @@ from datetime import datetime
 from enum import Enum
 from typing import Any, Callable, Dict, Optional
 
-from .constants import abort_key_prefix, default_queue_name, in_progress_key_prefix, \
-    job_key_prefix, result_key_prefix
+from .constants import abort_key_prefix, default_queue_name, in_progress_key_prefix, job_key_prefix, result_key_prefix
 from .utils import ms_to_datetime, poll, timestamp_ms
 
 logger = logging.getLogger('arq.jobs')
@@ -67,9 +66,14 @@ class Job:
         self._queue_name = _queue_name
         self._deserializer = _deserializer
 
-    async def abort(self, timeout: Optional[float] = None, *, pole_delay: float = 0.5) -> Any:
-        await self._redis.set('%s%s' % (abort_key_prefix, self.job_id), 1, expire=timeout)
-        return await self.result(timeout=timeout, pole_delay=pole_delay)
+    async def abort(
+        self, timeout: Optional[float] = None, *, pole_delay: float = 0.5, key_expire: Optional[int] = 10
+    ) -> Any:
+        await self._redis.set(f'{abort_key_prefix}{self.job_id}', b'1', expire=key_expire)
+        try:
+            await self.result(timeout=timeout, pole_delay=pole_delay)
+        except asyncio.CancelledError:
+            return True
 
     async def result(self, timeout: Optional[float] = None, *, pole_delay: float = 0.5) -> Any:
         """
--- a/arq/jobs.py
+++ b/arq/jobs.py
@@ -6,7 +6,8 @@ from datetime import datetime
 from enum import Enum
 from typing import Any, Callable, Dict, Optional
 
-from .constants import default_queue_name, in_progress_key_prefix, job_key_prefix, result_key_prefix
+from .constants import abort_key_prefix, default_queue_name, in_progress_key_prefix, \
+    job_key_prefix, result_key_prefix
 from .utils import ms_to_datetime, poll, timestamp_ms
 
 logger = logging.getLogger('arq.jobs')
@@ -65,6 +66,10 @@ class Job:
         self._redis = redis
         self._queue_name = _queue_name
         self._deserializer = _deserializer
+
+    async def abort(self, timeout: Optional[float] = None, *, pole_delay: float = 0.5) -> Any:
+        await self._redis.set('%s%s' % (abort_key_prefix, self.job_id), 1, expire=timeout)
+        return await self.result(timeout=timeout, pole_delay=pole_delay)
 
     async def result(self, timeout: Optional[float] = None, *, pole_delay: float = 0.5) -> Any:
         """
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -2,7 +2,6 @@ import asyncio
 import inspect
 import logging
 import signal
-from copy import copy
 from dataclasses import dataclass
 from datetime import datetime
 from functools import partial
@@ -172,6 +171,7 @@ class Worker:
         health_check_key: Optional[str] = None,
         ctx: Optional[Dict] = None,
         retry_jobs: bool = True,
+        abort_jobs: bool = False,
         max_burst_jobs: int = -1,
         job_serializer: Optional[Serializer] = None,
         job_deserializer: Optional[Deserializer] = None,
@@ -205,7 +205,6 @@ class Worker:
         else:
             self.redis_settings = None
         self.tasks = {}
-        self.abort_tasks = []
         self.main_task = None
         self.loop = asyncio.get_event_loop()
         self.ctx = ctx or {}
@@ -221,6 +220,8 @@ class Worker:
         self.on_stop = None
         # whether or not to retry jobs on Retry and CancelledError
         self.retry_jobs = retry_jobs
+        self.abort_jobs = abort_jobs
+        self._aborting_tasks = set()
         self.max_burst_jobs = max_burst_jobs
         self.job_serializer = job_serializer
         self.job_deserializer = job_deserializer
@@ -278,12 +279,26 @@ class Worker:
 
             if self.burst:
                 if 0 <= self.max_burst_jobs <= self._jobs_started():
-                    await asyncio.gather(*list(self.tasks.values()))
+                    await asyncio.gather(*self.tasks.values())
                     return
                 queued_jobs = await self.pool.zcard(self.queue_name)
                 if queued_jobs == 0:
-                    await asyncio.gather(*(self.tasks.values()))
+                    await asyncio.gather(*self.tasks.values())
                     return
+
+    async def _scan_abort_jobs(self):
+        todel, cur = [], b'0'
+        lk, mk = len(abort_key_prefix), f'{abort_key_prefix}*'
+        while cur:
+            cur, keys = await self.pool.scan(cur, match=mk)
+            for job_key in keys:
+                job_id = job_key[lk:]
+                if job_id not in self.tasks:
+                    todel.append(job_key)
+                    continue
+                self._aborting_tasks.add(job_id)
+        if todel:
+            await self.pool.delete(*todel)
 
     async def _poll_iteration(self):
         count = self.queue_read_limit
@@ -298,19 +313,18 @@ class Worker:
             job_ids = await self.pool.zrangebyscore(
                 self.queue_name, offset=self._queue_read_offset, count=count, max=now
             )
-            self.abort_tasks = [
-                k[len(abort_key_prefix):]
-                for k in await self.pool.keys('%s*' % abort_key_prefix)
-            ]
+
         await self.run_jobs(job_ids)
 
-        jobs = copy(self.tasks)
-        for job_id, t in jobs.items():
-            if t.done() :
+        if self.abort_jobs:
+            await self._scan_abort_jobs()
+
+        for job_id, t in list(self.tasks.items()):
+            if t.done():
                 del self.tasks[job_id]
                 # required to make sure errors in run_job get propagated
                 t.result()
-            elif job_id in self.abort_tasks:
+            elif job_id in self._aborting_tasks:
                 t.cancel()
 
         await self.heart_beat()
@@ -345,6 +359,7 @@ class Worker:
                     t = self.loop.create_task(self.run_job(job_id, score))
                     t.add_done_callback(lambda _: self.sem.release())
                     self.tasks[job_id] = t
+        print('end')
 
     async def run_job(self, job_id, score):  # noqa: C901
         start_ms = timestamp_ms()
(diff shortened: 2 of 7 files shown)


arq/jobs.py
```diff
@@ -6,7 +6,8 @@
 from enum import Enum
 from typing import Any, Callable, Dict, Optional
 
-from .constants import default_queue_name, in_progress_key_prefix, job_key_prefix, result_key_prefix
+from .constants import abort_key_prefix, default_queue_name, in_progress_key_prefix, \
```
samuelcolvin: I think this will fail linting, you need to run `make format`.

arq/jobs.py
```diff
@@ -66,6 +67,10 @@ def __init__(
         self._queue_name = _queue_name
         self._deserializer = _deserializer
 
+    async def abort(self, timeout: Optional[float] = None, *, pole_delay: float = 0.5) -> Any:
+        await self._redis.set('%s%s' % (abort_key_prefix, self.job_id), 1, expire=timeout)
```
samuelcolvin:
```suggestion
        await self._redis.set(f'{abort_key_prefix}{self.job_id}', b'1', expire=timeout)
```
Please use f-strings or if that's ugly, `.format()`

arq/worker.py
```diff
@@ -275,11 +278,11 @@ def run(self) -> None:
 
             if self.burst:
                 if 0 <= self.max_burst_jobs <= self._jobs_started():
-                    await asyncio.gather(*self.tasks)
+                    await asyncio.gather(*list(self.tasks.values()))
```
samuelcolvin:
```suggestion
                    await asyncio.gather(*self.tasks.values())
```
I doubt list is required here.

arq/worker.py
```diff
@@ -275,11 +278,11 @@ def run(self) -> None:
...
                     return
                 queued_jobs = await self.pool.zcard(self.queue_name)
                 if queued_jobs == 0:
-                    await asyncio.gather(*self.tasks)
+                    await asyncio.gather(*(self.tasks.values()))
```
samuelcolvin:
```suggestion
                    await asyncio.gather(*self.tasks.values())
```
These parentheses do nothing.

arq/worker.py
```diff
@@ -295,13 +298,20 @@ def run(self) -> None:
             job_ids = await self.pool.zrangebyscore(
                 self.queue_name, offset=self._queue_read_offset, count=count, max=now
             )
+            self.abort_tasks = [
```
samuelcolvin: Can we have a switch (default off) for whether we need to abort tasks, so this step can be omitted for those who don't need it. (this is a performance critical part of arq and `keys('...*')` is fairly slow I think.

Does `abort_tasks` need to be `self.abort_tasks`? I don't see why. Also gain, use f-strings.

Also, use a set not a list.

samuelcolvin: also, see [this](https://redis.io/commands/keys) warning, could we use, scan instead?

arq/worker.py
```diff
@@ -295,13 +298,20 @@ def run(self) -> None:
...
+                k[len(abort_key_prefix):]
+                for k in await self.pool.keys('%s*' % abort_key_prefix)
+            ]
         await self.run_jobs(job_ids)
 
-        for t in self.tasks:
-            if t.done():
-                self.tasks.remove(t)
+        jobs = copy(self.tasks)
```
samuelcolvin: this can be inlined, probably as `list(jobs.items())`

arq/worker.py
```diff
@@ -432,7 +442,7 @@ def run(self) -> None:
             logger.info('%6.2fs → %s(%s)%s', (start_ms - enqueue_time_ms) / 1000, ref, s, extra)
             # run repr(result) and extra inside try/except as they can raise exceptions
             try:
-                async with async_timeout.timeout(timeout_s):
+                async with async_timeout.timeout(timeout_s or None):
```
samuelcolvin: why is this required? Is it related to this PR?

arq/worker.py
```diff
@@ -450,9 +460,15 @@ def run(self) -> None:
...
-            elif self.retry_jobs and isinstance(e, asyncio.CancelledError):
-                logger.info('%6.2fs ↻ %s cancelled, will be run again', t, ref)
-                self.jobs_retried += 1
+            elif isinstance(e, asyncio.CancelledError):
+                if job_id in self.abort_tasks:
+                    logger.info('%6.2fs 🛇  %s aborted', t, ref)
+                    result = e
+                    finish = True
+                    self.abort_tasks.remove(job_id)
+                elif self.retry_jobs:
```
samuelcolvin: technically the `elif, else` logic has changed here.

I think it should be:
```suggestion
            elif isinstance(e, asyncio.CancelledError) and (self.retry_jobs or job_id in self.abort_tasks):
                if job_id in self.abort_tasks:
                    logger.info('%6.2fs 🛇  %s aborted', t, ref)
                    result = e
                    finish = True
                    self.abort_tasks.remove(job_id)
                else:
```

arq/worker.py
```diff
@@ -295,13 +298,20 @@ def run(self) -> None:
             job_ids = await self.pool.zrangebyscore(
                 self.queue_name, offset=self._queue_read_offset, count=count, max=now
             )
+            self.abort_tasks = [
+                k[len(abort_key_prefix):]
```
samuelcolvin: `len(abort_key_prefix)` can be a module level constant.

arq/worker.py
```diff
@@ -571,7 +592,7 @@ def handle_sig(self, signum):
             self.jobs_retried,
             len(self.tasks),
         )
-        for t in self.tasks:
+        for jid, t in self.tasks.items():
```
samuelcolvin:
```suggestion
        for t in self.tasks.values():
```
?

arq/worker.py
```diff
@@ -580,7 +601,7 @@ def handle_sig(self, signum):
     async def close(self):
         if not self.pool:
             return
-        await asyncio.gather(*self.tasks)
+        await asyncio.gather(*list(self.tasks.values()))
```
samuelcolvin: again, I think `list()` can be removed.

arq/worker.py
```diff
@@ -450,9 +460,15 @@ def run(self) -> None:
...
                 self.jobs_retried += 1
-            elif self.retry_jobs and isinstance(e, asyncio.CancelledError):
-                logger.info('%6.2fs ↻ %s cancelled, will be run again', t, ref)
-                self.jobs_retried += 1
+            elif isinstance(e, asyncio.CancelledError):
+                if job_id in self.abort_tasks:
+                    logger.info('%6.2fs 🛇  %s aborted', t, ref)
+                    result = e
+                    finish = True
+                    self.abort_tasks.remove(job_id)
```
samuelcolvin: we should delete the abort key here too.

arq/worker.py
david-e: I used in a task like this:
```python
async def cronjob(ctx, funcname, crontab, run_at_startup=False, *args, **kwargs):
    rds = ctx['redis']
    cj = CronTab(crontab)  # python-crontab string parser
    while True:
        dt = datetime.now()
        nr = cj.next(dt, default_utc=True)
        if nr or run_at_startup:
            run_at_startup = False
            await asyncio.sleep(nr)
        await rds.enqueue_job(funcname, *args, **kwargs)
 ```

and then define crontab in this way:

```python
func(cronjob, timeout=0)
```
If I pass None as timeout I cannot have the desired behaviour because of this, in worker.py:
```python
timeout_s = self.job_timeout_s if function.timeout_s is None else function.timeout_s
```
So I can schedule crontab remotely, start and stop them when needed, without the limit of timeout.

arq/worker.py
david-e: Sure, i changed a bit the code to use it more efficiently (I think). The results is something like that:

```python
    async def _abort_jobs(self):
        todel, cur = [], b'0'
        lk = len(abort_key_prefix)
        pool, tasks = self.pool, self.tasks
        while cur:
            cur, keys = await pool.scan(cur, match=f'{abort_key_prefix}*')
            for job_key in keys:
                job_id = job_key[lk:]
                if job_id not in tasks:
                    todel.append(job_key)
                    continue
                tasks[job_id].cancel()
                self._aborting_taks.add(job_id)
        if todel:
            await pool.delete(*todel)

    async def _poll_iteration(self):
        count = self.queue_read_limit
        if self.burst and self.max_burst_jobs >= 0:
            burst_jobs_remaining = self.max_burst_jobs - self._jobs_started()
            if burst_jobs_remaining < 1:
                return
            count = min(burst_jobs_remaining, count)

        async with self.sem:  # don't bother with zrangebyscore until we have "space" to run the jobs
            now = timestamp_ms()
            job_ids = await self.pool.zrangebyscore(
                self.queue_name, offset=self._queue_read_offset, count=count, max=now
            )
            if self.abort_jobs:
                await self._abort_jobs()
        await self.run_jobs(job_ids)

        for t in list(self.tasks.values()):
            if t.done():
                del self.tasks[job_id]
                # required to make sure errors in run_job get propagated
                t.result()

        await self.heart_beat()
```
What do you think?

arq/worker.py
david-e: ok

arq/worker.py
david-e: I fix it in this way:

```python
elif isinstance(e, asyncio.CancelledError) and job_id in self._aborting_tasks:
    logger.info('%6.2fs 🛇  %s aborted', t, ref)
    result = e
    finish = True
    self._aborting_tasks.remove(job_id)
   # redis key will be removed at next scan
elif isinstance(e, asyncio.CancelledError) and self.retry_jobs:
    logger.info('%6.2fs ↻ %s cancelled, will be run again', t, ref)
    self.jobs_retried += 1
```

samuelcolvin: I think this is less clear and duplicates the isinstance check.

david-e:
I have one question: why do you use polling every 0.5s (or custom interval) and not the publish/subscribe mechaniscm provided from redis?
subsribing to

samuelcolvin:
we could use publish and subscribe, but it would be a fair amount of work to setup.

The question is: why bother? Redis is amazingly performant, a few workers and clients polling it every 0.5s isn't going to have any effect on performance, even if reduced the timeout to 0.1s with 10 clients/workers, redis wouldn't notice in terms of throughput.

The only reason would be to reduce latency for jobs, responses. But if the worker is busy the polling delay won't have any effect as it'll make the request immediately when it's ready for more jobs.

In short, because I felt it wasn't worth. I'm prepared to be convinced otherwise, if you want to try, please create a new issue.

david-e:
> we could use publish and subscribe, but it would be a fair amount of work to setup.
> 
> The question is: why bother? Redis is amazingly performant, a few workers and clients polling it every 0.5s isn't going to have any effect on performance, even if reduced the timeout to 0.1s with 10 clients/workers, redis wouldn't notice in terms of throughput.
> 
> The only reason would be to reduce latency for jobs, responses. But if the worker is busy the polling delay won't have any effect as it'll make the request immediately when it's ready for more jobs.
> 
> In short, because I felt it wasn't worth. I'm prepared to be convinced otherwise, if you want to try, please create a new issue.

You are probably right. Publish/subscribe is the way I probably start with for a project like this, but these code works and it is really good, so I agree with you that the effort is a waste of time. I like the way you organize the project and the clearness of code. I hope to be helpful like you are for me.

arq/connections.py
```diff
@@ -101,6 +108,8 @@ async def enqueue_job(
         expires_ms = to_ms(_expires)
 
         with await self as conn:
+            if _del_prev_result:
+                await conn.delete(result_key_prefix + job_id)
```
samuelcolvin, line 112: why can't this be part of the pipelined operations?

samuelcolvin: why is this needed as part of this PR.

I think you may need to separate your work here into multiple pull requests for easier reveiwing.

arq/connections.py
```diff
@@ -133,34 +144,58 @@ async def enqueue_job(
                 return
         return Job(job_id, redis=self, _queue_name=_queue_name, _deserializer=self.job_deserializer)
 
-    async def _get_job_result(self, key) -> JobResult:
-        job_id = key[len(result_key_prefix) :]
+    async def _get_job_result(self, key, partial: bool = False) -> JobResult:
```
samuelcolvin, line 147: please can you explain the idea behind "partial" results - I think it's added since I last reviewed this PR?

Again, I think this might need to be in a separate PR.

arq/connections.py
```diff
@@ -133,34 +144,58 @@ async def enqueue_job(
...
+        return Job(job_id=job_id, redis=self, _queue_name=queue_name, _deserializer=self.job_deserializer)
+
     async def queued_jobs(self, *, queue_name: str = default_queue_name) -> List[JobDef]:
         """
         Get information about queued, mostly useful when testing.
         """
         jobs = await self.zrange(queue_name, withscores=True)
         return await asyncio.gather(*[self._get_job_def(job_id, score) for job_id, score in jobs])
 
+    async def all_jobs_queued(self, *, queue_name: str = default_queue_name) -> List[JobDef]:
```
samuelcolvin, line 184: why is this method required?

Again I don't think it's related to this PR.

arq/connections.py
```diff
@@ -133,34 +144,58 @@ async def enqueue_job(
...
+        """
+        Get information about queued jobs, same of queued_jobs, but adopting naming convention.
+        """
+        return await self.queued_jobs(queue_name=queue_name)
+
+    async def all_jobs_running(self) -> List[JobDef]:
```
samuelcolvin, line 190: same

arq/connections.py
```diff
@@ -133,34 +144,58 @@ async def enqueue_job(
...
+    async def _get_job_result(self, key, partial: bool = False) -> JobResult:
+        job_id = key[len(result_key_prefix if not partial else partial_result_key_prefix) :]
         job = Job(job_id, self, _deserializer=self.job_deserializer)
-        r = await job.result_info()
+        r = await job.result_info(partial)
         r.job_id = job_id
         return r
 
-    async def all_job_results(self) -> List[JobResult]:
+    async def all_jobs_results(self) -> List[JobResult]:
```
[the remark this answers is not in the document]
david-e: Sorry, I continue to develop my changes on the same branch. Now I revert it for the only purpose of this PR, moving other stuff in a personal branch.

arq/jobs.py
```diff
@@ -66,6 +66,15 @@ def __init__(
         self._queue_name = _queue_name
         self._deserializer = _deserializer
 
+    async def abort(
+        self, timeout: Optional[float] = None, *, pole_delay: float = 0.5, key_expire: Optional[int] = 10
+    ) -> Any:
```
samuelcolvin, line 71: I'm not clear on the return type here - what does it mean? I think it should be `bool`

arq/worker.py
samuelcolvin: please remove this.

docs/index.rst
```diff
@@ -167,6 +167,10 @@ optionally with a duration to defer rerunning the jobs by:
 
 .. literalinclude:: examples/retry.py
 
+It is also possible to abort long running tasks, passing ``abort_jobs=True`` to ``WorkerSettins``.
```
samuelcolvin, line 170:
```suggestion
It is also possible to abort long running tasks, passing ``abort_jobs=True`` to ``WorkerSettings``.
```

arq/worker.py
```diff
@@ -275,13 +279,27 @@ async def main(self):
...
+        todel, cur = [], b'0'
+        lk, mk = len(abort_key_prefix), f'{abort_key_prefix}*'
+        while cur:
+            cur, keys = await self.pool.scan(cur, match=mk)
+            for job_key in keys:
+                job_id = job_key[lk:]
+                if job_id not in self.tasks:
+                    todel.append(job_key)
+                    continue
+                self._aborting_tasks.add(job_id)
```
samuelcolvin, line 299:
```suggestion
                if job_id in self.tasks:
                    self._aborting_tasks.add(job_id)
                else:
                    todel.append(job_key)
```
slightly cleaner.

samuelcolvin:
Closing this as inactive, but I can re-open or review a new PR if a fix becomes available.

PR #177, opened 2020-04-18
Make password setting Optional[str] for RedisSettings

elatt:
`None` is a valid value for the password prop in `RedisSettings`


samuelcolvin:
thanks so much.

PR #183, opened 2020-04-25
Option to disable registration of signal handlers

### LINKED ISSUE #182: Running arq inside another async app
I want to run arq from within another app that controls the asyncio event loop (FastAPI).

I've been able to start it up and get it working, but the issue is on shutdown where arq catches the sigint/sigterm signals and these never get passed up. This makes the process hang in a dead state.

```
class Worker:
    def __init__(self):
        pass

    async def start(self):
        # loop = asyncio.get_event_loop()
        self.worker = create_worker(worker_settings)
        # loop.remove_signal_handler(SIGINT)
        # loop.remove_signal_handler(SIGTERM)
        self.task = asyncio.create_task(self.worker.async_run())

    async def close(self):
        logger.debug("Starting worker close")
        await self.worker.close()
        logger.debug("Ending worker close")
```

I've found that if I remove the loop's signal handlers that arq created I can exit, but this throws errors since no signal handlers are now present.

Signal handlers are added here:
https://github.com/samuelcolvin/arq/blob/master/arq/worker.py#L215
```
        self._add_signal_handler(signal.SIGINT, self.handle_sig)
        self._add_signal_handler(signal.SIGTERM, self.handle_sig)
```

In the app startup/shutdown I will startup/shutdown arq as necessary, but would need an option to pass in to prevent arq from registering its own handlers.

kdorsel:
Used when you want to listen to signal handlers on your own

Closes #182

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,7 @@ History
 v0.19.1 (unreleased)
 ....................
 * fix timestamp issue in _defer_until without timezone offset, #182
+* add option to disable signal handler registration for running inside other frameworks, #183
 
 v0.19.0 (2020-04-24)
 ....................
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,7 +6,7 @@ History
 v0.19.1 (unreleased)
 ....................
 * fix timestamp issue in _defer_until without timezone offset, #182
-* add option to disable signal handler registration for running inside other frameworks, #183
+* add option to disable signal handler registration from running inside other frameworks, #183
 
 v0.19.0 (2020-04-24)
 ....................
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -133,7 +133,7 @@ class Worker:
     :param burst: whether to stop the worker once all jobs have been run
     :param on_startup: coroutine function to run at startup
     :param on_shutdown: coroutine function to run at shutdown
-    :param no_sig: does not register signal handlers, used when running inside other async framework
+    :param sigs: default true, register signal handlers, set to false when running inside other async framework
     :param max_jobs: maximum number of jobs to run at a time
     :param job_timeout: default job timeout (max run time)
     :param keep_result: default duration to keep job results for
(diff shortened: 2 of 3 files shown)


arq/worker.py
```diff
@@ -133,6 +133,7 @@ class Worker:
     :param burst: whether to stop the worker once all jobs have been run
     :param on_startup: coroutine function to run at startup
     :param on_shutdown: coroutine function to run at shutdown
+    :param no_sig: does not register signal handlers, used when running inside other async framework
```
samuelcolvin, line 136:
```suggestion
    :param sigs: default True, whether register signal handlers, set to False when running inside other async framework
```
I don't like double negatives `not self._no_sig`, even when the new `False` mode is a change.

Tinche:
We already do this by subclassing the Worker class and overriding `_add_signal_handler` to a NOOP, but a parameter would be better from a usability perspective.

Since this is asyncio, running Arq in the same process as your f.e. http handlers is a big win. Instead of 3 deployments (web handlers, workers, cron) you can just run one, and scaling is much easier.

Why is the parameter typed as `Optional[bool]` though, it can just be `bool`?

arq/worker.py
```diff
@@ -159,6 +160,7 @@ def __init__(
         burst: bool = False,
         on_startup: Optional['StartupShutdown'] = None,
         on_shutdown: Optional['StartupShutdown'] = None,
+        sigs: bool = True,
```
samuelcolvin, line 163:
```suggestion
        handle_signals: bool = True,
```
Bit clearer for those new to the project.

HISTORY.rst
```diff
@@ -6,6 +6,7 @@ History
 v0.19.1 (unreleased)
 ....................
 * fix timestamp issue in _defer_until without timezone offset, #182
+* add option to disable signal handler registration for running inside other frameworks, #183
```
samuelcolvin, line 9:
```suggestion
* add option to disable signal handler registration from running inside other frameworks, #183
```

arq/worker.py
```diff
@@ -133,6 +133,7 @@ class Worker:
     :param burst: whether to stop the worker once all jobs have been run
     :param on_startup: coroutine function to run at startup
     :param on_shutdown: coroutine function to run at shutdown
+    :param sigs: default true, register signal handlers, set to false when running inside other async framework
```
samuelcolvin, line 136:
```suggestion
    :param handle_signals: default true, register signal handlers, set to false when running inside other async framework
```

samuelcolvin:
thank you so much.

PR #187, opened 2020-05-06
windows event loop sig handler

jessekrubin:
except not implemented error for windows when adding sig handler to event loop

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -578,7 +578,7 @@ class Worker:
     def _add_signal_handler(self, signum: Signals, handler: Callable[[Signals], None]) -> None:
         try:
             self.loop.add_signal_handler(signum, partial(handler, signum))
-        except NotImplementedError:
+        except NotImplementedError:  # pragma: no cover
             logger.debug('Windows does not support adding a signal handler to an eventloop')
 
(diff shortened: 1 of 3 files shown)


arq/worker.py
```diff
@@ -576,7 +576,10 @@ async def record_health(self) -> None:
             self._last_health_check_log = log_suffix
 
     def _add_signal_handler(self, signum: Signals, handler: Callable[[Signals], None]) -> None:
-        self.loop.add_signal_handler(signum, partial(handler, signum))
+        try:
+            self.loop.add_signal_handler(signum, partial(handler, signum))
+        except NotImplementedError:
+            logger.debug('Oy vey; Windows does not support adding a signal handler to an eventloop')
```
samuelcolvin, line 582:
```suggestion
            logger.debug('Windows does not support adding a signal handler to an eventloop')
```

Bit non standard and confusing.

jessekrubin: Solid. 

PS: windows regularly drives me bananas

samuelcolvin:
thanks so much.

jessekrubin:
> thanks so much.

No problemo @samuelcolvin

PR #191, opened 2020-05-21
Introduce default queue name to ArqRedis

Tinche:
When using a single central Redis with multiple services, it's necessary to use different queues.

It's cumbersome to have to pass the queue name when enqueing jobs all the time, so we add the ability for an instance of ArqRedis to remember it.

We also make the worker be able to get the default queue name from ArqRedis, to not have to repeat ourselves.

This should be backwards compatible.

I've also fixed some timezone issues which were failing tests on my machine, which is not in UTC.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,8 @@ History
 v0.19.1 (unreleased)
 ....................
 * fix timestamp issue in _defer_until without timezone offset, #182
+* Add ``default_queue_name`` to ``create_redis_pool`` and ``ArqRedis``.
+* ``Worker`` can retrieve the ``queue_name`` from the connection pool, if present.
 
 v0.19.0 (2020-04-24)
 ....................
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,8 +6,8 @@ History
 v0.19.1 (unreleased)
 ....................
 * fix timestamp issue in _defer_until without timezone offset, #182
-* Add ``default_queue_name`` to ``create_redis_pool`` and ``ArqRedis``.
-* ``Worker`` can retrieve the ``queue_name`` from the connection pool, if present.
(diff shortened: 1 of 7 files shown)


arq/worker.py
```diff
@@ -174,6 +174,11 @@ def __init__(
         job_deserializer: Optional[Deserializer] = None,
     ):
         self.functions: Dict[str, Union[Function, CronJob]] = {f.name: f for f in map(func, functions)}
+        if queue_name is None:
+            if redis_pool is not None:
+                queue_name = redis_pool.default_queue_name
+            else:
+                raise Exception('If queue_name is absent, redis_pool must be present.')
```
samuelcolvin, line 181:
```suggestion
                raise ValueError('If queue_name is absent, redis_pool must be present.')
```

HISTORY.rst
```diff
@@ -6,6 +6,8 @@ History
 v0.19.1 (unreleased)
 ....................
 * fix timestamp issue in _defer_until without timezone offset, #182
+* Add ``default_queue_name`` to ``create_redis_pool`` and ``ArqRedis``.
+* ``Worker`` can retrieve the ``queue_name`` from the connection pool, if present.
```
samuelcolvin, line 10:
```suggestion
* Add ``default_queue_name`` to ``create_redis_pool`` and ``ArqRedis``, #191
* ``Worker`` can retrieve the ``queue_name`` from the connection pool, if present
```

Tinche:
@samuelcolvin thanks for looking at this. Just FYI, I have a draft blog post about Arq in the pipeline, and ideas for a few more PRs.

samuelcolvin:
> I have a draft blog post about Arq in the pipeline, and ideas for a few more PRs.

Great, contributions welcome. I'm pretty busy at the moment but will try my best to respond to PRs promptly.

PR #194, opened 2020-06-27
Remove gather

Tinche:
`asyncio.gather` makes no guarantees about the order of execution of the coroutines it's given, and in this case ordering matters (the unwatch needs to execute before the watch, for example). So I've explicitly broken out the gather into individual awaits, then wrapped the awaits in a pipeline for less I/O.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,9 +5,10 @@ History
 
 v0.19.1 (unreleased)
 ....................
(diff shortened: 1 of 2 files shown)


jessekrubin:
👍 I support this (fwiw)

samuelcolvin:
looks good, good catch, can you add an entry to `HISTORY`.

Issue #161, opened 2019-11-02
Include queue_name in job result


pratheekrebala:
Is it possible to include queue_name in the job result object? It would be really helpful to trace job failures back to the worker if queue_name was included in the result object..

samuelcolvin:
Fine by me.

PR #198, opened 2020-07-29
Added queue_name attribute to JobResult

ccharlesgb:
This pull request addresses issue #161 

I didn't feel like I needed to add any new tests as modifying the existing tests checks that the attribute exists in the result.

Also technically I could add the queue_name attribute in the JobDef as well but I don't really see a use for that yet but happy to add it there as well.

### DIFF
--- a/arq/jobs.py
+++ b/arq/jobs.py
@@ -50,6 +50,7 @@ class JobResult(JobDef):
     result: Any
     start_time: datetime
     finish_time: datetime
+    queue_name: str
     job_id: Optional[str] = None
 
 
@@ -172,6 +173,7 @@ def serialize_result(
     start_ms: int,
     finished_ms: int,
     ref: str,
+    queue_name: str,
     *,
     serializer: Optional[Serializer] = None,
 ) -> Optional[bytes]:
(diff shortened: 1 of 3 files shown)


arq/jobs.py
```diff
@@ -247,6 +250,7 @@ def deserialize_result(r: bytes, *, deserializer: Optional[Deserializer] = None)
             result=d['r'],
             start_time=ms_to_datetime(d['st']),
             finish_time=ms_to_datetime(d['ft']),
+            queue_name=d['q'],
```
ccharlesgb, line 253: I was slightly concerned that this change will break compatibility if the worker is running an older version of arq if this change goes through then it would cause this to raise a KeyError. Am I right in this thinking and should I change this to `d.get('q')` ?

ccharlesgb:
@samuelcolvin Sorry to nudge. But would be great to get your thoughts on this

arq/jobs.py
samuelcolvin: yes I think you're right.

Since it's been so long, I'll fix this myself now.

samuelcolvin:
So sorry for the massive delay on this, thank you so much for your contribution.

I've made the change you suggested and merged this. I'm working on a new release now.

Issue #196, opened 2020-07-01
Unique cron executes multiple times when running multiple workers


Tinche:
Hi,

we're seeing this in our services. I believe the problem can be summarized as follows.

Poll iterations are done every 500 ms. This is a long time, relatively speaking.

Worker 0, at t0:
0) Poll the queue, enqueue the cron.

Worker 1, at t0 + 10ms:
1) Poll the queue, get the cron job ID
2) Check in_progress_key, queue score. Start the job
3) Job finishes really quickly (~1 ms)
4) The cleanup atomically deletes the in_progress_key, retry key, job key, queue score

Worker 2, at t0 + 20ms:
5) Poll the queue, schedule the cron

Worker 3, at t0 + 30ms:
6) Poll the queue, get the cron job ID
7) Check in_progress_key, queue score. Everything is in order since the last execution cleaned up after itself. Start the job

I'm willing to contribute a fix for this, but let's discuss it first.

We should probably have a way of storing the fact that a cron job finished executing successfully. (I'm not sure how retries are handled for unique crons, so I'm ignoring that case right now.) The simplest way of doing this would be to not delete the in_progress key, but set it to expire in a number of seconds. (The number should be greater than the poll duration, and should probably be greater than average clock drift if the servers don't have NTP configured. Let's say 60 seconds.)

I think this change alone would fix this issue. To be a little more efficient though, we could also check for the existence of this key while enqueueing the cron, so not every worker process enqueues it separately, to save on a little work. Since this is an optimization, it doesn't have to be perfectly free of race conditions.

havardthom:
Hey, also ran into this issue when running a unique cron job with 4 arq workers (job runs two times in my case). Seems like there is a pull request ready to fix this issue, any plans on merging it?

PR #200, opened 2020-08-22
Cron refactor

Tinche:
This is an attempt of fixing #196.

Instead of enqueueing a cron job after the cron time, we enqueue before.

This makes the cron job run closer to its schedule, and the enqueues are idempotent (it's the same if one worker enqueues or 20).

A more detailed explanation follows.

We introduce the concept of a cron scheduling window, that's tied to the polling delay (the polling delay default is 0.5s), except the cron scheduling window length is clamped at a minimum value of 0.5s. Anyway, let's assume the poll delay is 0.5s, and hence the cron window length is 0.5s.

We start keeping track of the exact time of the last heartbeat.

On each heartbeat, we iterate over the defined cronjobs, choose cronjobs that are inside the window and enqueue them. The window is defined as starting at (last_heartbeat_time + cron window length), and ending at (now + cron window length).

Also, for cronjobs that finish successfully, we keep the in-progress key in Redis for 1 second (instead of deleting it outright). This is to stop the cronjob being enqueued again by workers with skewed clocks, and to handle edge cases. (We might want to use a larger buffer than 1s.)

### DIFF
--- a/arq/constants.py
+++ b/arq/constants.py
@@ -4,3 +4,4 @@ in_progress_key_prefix = 'arq:in-progress:'
 result_key_prefix = 'arq:result:'
 retry_key_prefix = 'arq:retry:'
 health_check_key_suffix = ':health-check'
+keep_cronjob_progress = 1.0
--- a/arq/constants.py
+++ b/arq/constants.py
@@ -4,4 +4,6 @@ in_progress_key_prefix = 'arq:in-progress:'
 result_key_prefix = 'arq:result:'
 retry_key_prefix = 'arq:retry:'
 health_check_key_suffix = ':health-check'
-keep_cronjob_progress = 1.0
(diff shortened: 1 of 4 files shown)


Tinche:
I simplified the cron scheduling - we don't need to keep track of the last heartbeat. The state in the actual cronjob objects is enough.

samuelcolvin:
I've made some small changes here, @Tinche could you review and confirm you're happy with this.

I'll try and deploy towards the end of the week.

Tinche:
@samuelcolvin looks great! I see there's a conflict in worker.py, want me to rebase and fix it?

samuelcolvin:
Great, please do if you have the time.

Tinche:
Rebased!

samuelcolvin:
Thanks so much for this.

PR #202, opened 2020-08-30
Add helper to create RedisSettings instance given a well-formed connection URL

inactivist:
This PR provides a new `arq.connections.RedisSettings` class method, `from_redis_url()`, that returns a new instance from a parsed redis URL.


arq/connections.py
```diff
@@ -38,6 +39,19 @@ class RedisSettings:
     sentinel: bool = False
     sentinel_master: str = 'mymaster'
 
+    @classmethod
+    def from_redis_url(cls, url: str):
```
samuelcolvin, line 43: Please set the return type.

arq/connections.py
```diff
@@ -38,6 +39,19 @@ class RedisSettings:
...
+        """Parse a redis: connection URI, return a new
+        arq.connections.RedisSettings instance
+
+        :param url: The redis: url to parse
+        :return: :class:`arq.connections.RedisSettings` instance
+        """
+        address, options = parse_url(url)
```
samuelcolvin, line 50: Does this check the URL scheme is correct?

inactivist: IIRC aioredis [`parse_url`](https://github.com/aio-libs/aioredis/blob/master/aioredis/util.py#L148-L149) does some error checking via asserts.

samuelcolvin: Ye, aioredis has a lot of problems, I've started a new redis connection library, but it's far from finished.

I guess `assert` is okay for now here, very few people actually bother to disable debug in python.

arq/connections.py
```diff
@@ -38,6 +39,21 @@ class RedisSettings:
...
+
+        :param url: The redis: url to parse
+        :return: :class:`arq.connections.RedisSettings` instance
+        :raises: AssertionError: if url is malformed
+            NOTE: Validation performed by `aioredis.util.parse_url` via
+            `assert` statements, and thus no validation takes place if various
+            Python interpreter optimizations are enabled at runtime.
+        """
+        address, options = parse_url(url)
+        return cls(host=address[0], port=address[1], password=options.get('password'))
```
samuelcolvin, line 55: you should be able to set the database here too. see [here](https://github.com/samuelcolvin/foxglove/blob/22723a32b3a2c7ba8db24af8c7f2a3d0bb218b10/foxglove/settings.py#L162)

inactivist: I'm adding a kwargs passthru to `RedisSettings` constructor for remaining args (see subsequent commits)

arq/connections.py
inactivist: > I guess `assert` is okay for now here, very few people actually bother to disable debug in python.

Truth be told, as a user of arq, I wouldn't expect to perform validation on the redis connection URL string, other than to raise exceptions if grossly malformed (IOW can't parse in our helper code)

If anything, I assume it's the responsibility of underlying redis connection library, let it raise an exception if it wants to perform this level of validation.

inactivist:
BTW and OT: contributor docs describing development env setup and workflow would be helpful.  I'm still trying to figure out how to configure mypy to pass cleanly locally even though the CI server here seems happy.

```
~/src/other/arq$ make
flake8 arq/ tests/
isort -rc arq tests --check-only -df
black -S -l 120 --target-version py37 arq tests --check
All done! ✨ 🍰 ✨
20 files would be left unchanged.
mypy arq
arq/cron.py:7: error: Cannot find implementation or library stub for module named 'pydantic.utils'
arq/worker.py:12: error: Cannot find implementation or library stub for module named 'async_timeout'
arq/worker.py:14: error: Cannot find implementation or library stub for module named 'pydantic.utils'
arq/cli.py:9: error: Cannot find implementation or library stub for module named 'pydantic.utils'
arq/cli.py:9: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
Found 4 errors in 3 files (checked 11 source files)
Makefile:33: recipe for target 'mypy' failed
make: *** [mypy] Error 1

arq/connections.py
inactivist: > you should be able to set the database here too

I trust I've addressed your concern here (please correct me if I'm wrong).  So I'm gong to resolve this conversation.

inactivist:
Closing, as it appears this use case has been addressed in a recent update.

PR #203, opened 2020-09-02
Fix issue where pool serializer would not be same as worker

lafka:
The issue was encountered when initializing redis through RedisSettings
instead of explicitly setting the arq connection pool.

If cron was used and `job_serializer` was not `pickle` the cron item would be
serialized with pickle.dumps and then attempted to deserialize with the configured
`job_deserializer` in worker, causing an error.

Quick workaround is to create the pool yourself, otherwise always use the
pickle serializer.

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -276,7 +276,9 @@ class Worker:
 
     async def main(self) -> None:
         if self._pool is None:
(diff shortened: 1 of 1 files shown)


samuelcolvin:
great, thank you for the contribution.

PR #207, opened 2020-09-24
Fix coroutine name in docs

facundojmaero:
Just a small fix in the start of the tutorial


samuelcolvin:
Good spot, thank you. 🙏

PR #209, opened 2020-10-06
the original aioredis supports UNIX socket why not us?

ihakh:
the pure string in aioredis address is for UNIX socket connection
this way we do not drop the support for it


samuelcolvin:
I don't think we can make this change as it is now, since I think it will be backwards incompatible with many people's usage of `RedisSettings`.

Perhaps instead we could add `socket_address` or similar with overrides `(host, port)`?

samuelcolvin:
Closing this as no response, happy to accept a new PR to add a new `socket_address` address argument to `RedisSettings`

PR #212, opened 2020-10-21
Add method to cancel jobs

sgranade:
Added method to cancel jobs through a new `abort()` function on Job. This allows long-running or deferred jobs to be cancelled.

This is a cleanup of PR #173 , since the submitter didn't finish the requested changes, with one design change: a `Worker` will cancel aborted jobs by default. I made the design change so that people aren't surprised when they call `abort()` on a `Job` and nothing happens. My guess is that it's more common to want to allow cancelling jobs than to block that operation.

This addresses issue #172 .

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -332,12 +332,12 @@ class Worker:
             await self.sem.acquire()
             in_progress_key = in_progress_key_prefix + job_id
             with await self.pool as conn:
-                _, _, ongoing_exists, score = await asyncio.gather(
-                    conn.unwatch(),
-                    conn.watch(in_progress_key),
-                    conn.exists(in_progress_key),
-                    conn.zscore(self.queue_name, job_id),
-                )
+                pipe = conn.pipeline()
+                pipe.unwatch()
+                pipe.watch(in_progress_key)
+                pipe.exists(in_progress_key)
+                pipe.zscore(self.queue_name, job_id)
+                _, _, ongoing_exists, score = await pipe.execute()
                 if ongoing_exists or not score:
                     # job already started elsewhere, or already finished and removed from queue
                     self.sem.release()
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -147,6 +147,7 @@ class Worker:
     :param health_check_key: redis key under which health check is set
     :param ctx: dictionary to hold extra user defined state
     :param retry_jobs: whether to retry jobs on Retry or CancelledError or not
+    :param abort_jobs: whether to cancel jobs on a call to :func:`arq.jobs.Job.abort`
     :param max_burst_jobs: the maximum number of jobs to process in burst mode (disabled with negative values)
     :param job_serializer: a function that serializes Python objects to bytes, defaults to pickle.dumps
     :param job_deserializer: a function that deserializes bytes into Python objects, defaults to pickle.loads
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -175,7 +175,7 @@ class Worker:
         health_check_key: Optional[str] = None,
         ctx: Optional[Dict[Any, Any]] = None,
         retry_jobs: bool = True,
-        abort_jobs: bool = False,
+        abort_jobs: bool = True,
         max_burst_jobs: int = -1,
         job_serializer: Optional[Serializer] = None,
         job_deserializer: Optional[Deserializer] = None,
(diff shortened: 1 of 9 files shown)


sgranade:
Thinking more about this, the `abort()` function should return the job's value if it completed before `abort()` was called. Otherwise, we lose the job's returned value, which can be useful to have.

I'm wondering if it makes sense for `abort()` to return the value if the job completed before being cancelled, and then either raising a `CancelledError` or returning a sentinel value if the job aborted as planned.

I'm happy to make either change to the design and re-submit the PR.

arq/worker.py
```diff
@@ -316,14 +321,35 @@ async def _poll_iteration(self) -> None:
...
-                self.tasks.remove(t)
+                del self.tasks[job_id]
                 # required to make sure errors in run_job get propagated
                 t.result()
+            elif job_id in self._aborting_tasks:
+                t.cancel()
 
         await self.heart_beat()
 
+    async def _scan_abort_jobs(self) -> None:
```
samuelcolvin, line 337: I know this is an internal method, but I think we still need a docstring explaining what's going on. It's not at all trivial.

In particular I'm not clear what `orphaned_job_keys` is or why it could happen.

samuelcolvin: if `orphaned_job_keys` is just about clearing up keys added via `abort()` that don't apply here, I think cleaner to just omit it, those keys should expire soon anyway.

arq/worker.py
```diff
@@ -475,6 +501,12 @@ async def job_failed(exc: Exception) -> None:
                 if e.defer_score:
                     incr_score = e.defer_score + (timestamp_ms() - score)
                 self.jobs_retried += 1
+            elif job_id in self._aborting_tasks and isinstance(e, asyncio.CancelledError):
+                logger.info('%6.2fs 🛇  %s aborted', t, ref)
```
samuelcolvin, line 505: The symbol here doesn't work for me on mac (I can't see it in the browser in github, in pycharm or in the terminal)

How about using the "CIRCLED DIVISION SLASH" symbol ⊘?

That works nicely everywhere and [apparently](https://en.wikipedia.org/wiki/No_symbol#Unicode_Combining_Character) has good font support.

image

arq/worker.py
```diff
@@ -459,14 +485,14 @@ async def job_failed(exc: Exception) -> None:
...
-            except Exception as e:
+            except (Exception, asyncio.CancelledError) as e:
                 exc_extra = getattr(e, 'extra', None)
                 if callable(exc_extra):
                     exc_extra = exc_extra()
                 raise
             else:
                 result_str = '' if result is None else truncate(repr(result))
-        except Exception as e:
+        except (Exception, asyncio.CancelledError) as e:
```
samuelcolvin, line 495: this is good, but it's important to note that it fixes current behaviour and will mean jobs are retried correctly upon restarting, not just by default when the `in_progress` key expires.

docs/examples/job_abort.py
```diff
@@ -0,0 +1,60 @@
...
+from arq.connections import RedisSettings
+# requires `pip install devtools`, used for pretty printing of job info
+from devtools import debug
+
+
+async def the_task(ctx):
+    print('running the task')
+    return 42
+
+async def cron_the_task(ctx, crontab, funcname):
```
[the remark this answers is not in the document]
samuelcolvin: Try to make the example as simple as possible, something like 

```py
import asyncio
from arq import create_pool


async def do_stuff(ctx):
    print('doing stuff...')
    await asyncio.sleep(10)
    return f'done {x}'


async def main():
    job = await redis.enqueue_job('do_stuff')
    await asyncio.sleep(1)
    await job.abort()


class WorkerSettings:
    functions = [do_stuff]


if __name__ == '__main__':
    asyncio.run(main())
```

samuelcolvin (commented):
otherwise this is looking good.

I know it's been a long time, so if you're not able to fix it over the next few days. I'll fix it for you.

samuelcolvin:
I've re-written this significantly, in particular:
* I'm now using a sorted set, not a namespace to store IDs of jobs to abort, this is for performance reasons
* I'm now cancelling the actual job execution not the entire `run_job` task
* I've added logic to "abort" a job before it begins

**Side note:** It's lucky I dug into this since as it was before, if you tried to abort a job before it began, the worker itself stopped silently!

This is because cancelling `run_job` when it **wasn't** on the `result = await function.coroutine(ctx, *args, **kwargs)` caused a `CancelledError` to propagate which in the end is silently handled by 

https://github.com/samuelcolvin/arq/blob/8996603c7a45410f227d69a0a2b101b4632b072e/arq/worker.py#L249-L251

samuelcolvin:
Thanks so much @sgranade, I'll release this soon.

Issue #190, opened 2020-05-18
Job cancellation/retry is broken


Kamforka:
I've tried to reproduce the automatic job retry mechanism of `arq` based on [Retrying jobs and cancellation](https://arq-docs.helpmanual.io/#retrying-jobs-and-cancellation).

However whenever I press Ctrl+C  - once my task already started but before it's finished - the worker produces a nasty exception traceback like this:

```python3
13:14:31: Starting worker for 1 functions: long_running_task
13:14:31: redis_version=5.0.9 mem_usage=858.92K clients_connected=2 db_keys=15
13:14:33:   0.24s → caefef46a4fa43618ece3fd77f92cd2a:long_running_task()
Running...
^C13:14:34: shutdown on SIGINT ◆ 0 jobs complete ◆ 0 failed ◆ 0 retries ◆ 1 ongoing to cancel
Traceback (most recent call last):
  File "/home/s6dnau/git/hx-api/venv/bin/arq", line 10, in <module>
    sys.exit(cli())
  File "/home/s6dnau/git/hx-api/venv/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/s6dnau/git/hx-api/venv/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/s6dnau/git/hx-api/venv/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/s6dnau/git/hx-api/venv/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/s6dnau/git/hx-api/venv/lib/python3.8/site-packages/arq/cli.py", line 48, in cli
    run_worker(worker_settings_, **kwargs)
  File "/home/s6dnau/git/hx-api/venv/lib/python3.8/site-packages/arq/worker.py", line 630, in run_worker
    worker.run()
  File "/home/s6dnau/git/hx-api/venv/lib/python3.8/site-packages/arq/worker.py", line 235, in run
    self.loop.run_until_complete(self.close())
  File "/home/s6dnau/.pyenv/versions/3.8.0/lib/python3.8/asyncio/base_events.py", line 608, in run_until_complete
    return future.result()
asyncio.exceptions.CancelledError
```

Then I immediately restart my worker and the cancelled job is only retried after ~310 seconds:

[...]

PR #213, opened 2020-10-22
Fix Worker and Job to catch CancelledException.

sgranade:
As issue #190 noted, `except Exception as e` doesn't catch an `asyncio.CancelledError`. Some of the code in `Worker` clearly expects `asyncio.CancelledError` to be caught properly if raised in a `Job`. This PR fixes the appropriate lines to catch the error properly instead of kicking it all the way out to the enclosing Worker task, which then gets cancelled.

### DIFF
--- a/arq/jobs.py
+++ b/arq/jobs.py
@@ -86,7 +86,7 @@ class Job:
                 result = info.result
                 if info.success:
                     return result
-                elif isinstance(result, Exception) or isinstance(result, asyncio.CancelledError):
+                elif isinstance(result, (Exception, asyncio.CancelledError)):
                     raise result
                 else:
                     raise SerializationError(result)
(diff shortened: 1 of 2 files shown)


arq/jobs.py
```diff
@@ -86,7 +86,7 @@ async def result(self, timeout: Optional[float] = None, *, pole_delay: float = 0
                 result = info.result
                 if info.success:
                     return result
-                elif isinstance(result, Exception):
+                elif isinstance(result, Exception) or isinstance(result, asyncio.CancelledError):
```
kornicameister, line 89: Not a maintainer, but wouldn't: 
```suggestion
                elif isinstance(result, (Exception, asyncio.CancelledError)):
```
work too ?

samuelcolvin:
This correct but is also covered by #212

samuelcolvin:
I've changed my mind, I'm going to use this solution since I think #212 might not be merged before the next release.

Thanks for fixing this.

PR #215, opened 2020-11-24
Fix to allow custom serializers

TimOrme:
This fixes an issue where custom serializers are not used if the job is enqueued from the ctx within an existing task.

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -276,9 +276,9 @@ class Worker:
 
     async def main(self) -> None:
         if self._pool is None:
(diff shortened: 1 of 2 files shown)


explorigin:
This is the same basic change as in https://github.com/samuelcolvin/arq/pull/203 but has additional tests.

TimOrme:
> This is the same basic change as in #203 but has additional tests.

Ah didn't see that one, thanks for linking.  I'll can leave this one here in case the tests are valuable?

samuelcolvin:
Thanks for the contribution.

PR #226, opened 2021-02-06
Use asyncio.run() in examples

cmangla:
Avoids the `Task was destroyed but it is pending!` error message. Looks a little neater too.


samuelcolvin:
great, thank you so much.

Issue #230, opened 2021-02-16
`job_timeout` not working


cmangla:
In the following snippet, I try to create a job that should time out:
```python3
import time
import asyncio
from arq import create_pool
from arq.connections import RedisSettings

async def thetask(ctx, tm):
    start_time = time.time()
    time.sleep(tm) #(*1)
    return time.time() - start_time

async def main():
    rq = await create_pool(RedisSettings())
    job = await rq.enqueue_job('thetask', 30) #(*2)
    print(await job.result())

class WorkerSettings:
    functions = [thetask]
    job_timeout = 10 #(*3)

if __name__ == '__main__':
    asyncio.run(main())
```

When run using arq, the task takes up the full 30 seconds (as specified in line marked `*2`)  instead of timing out in 10 (as set in `*3`). Line `*1` is deliberately synchronous, as are my workloads where I'm seeing this issue, which are cpu-intensive subprocesses.

If I replace `thetask()` above with asynchronous code like so:
```Python3
async def thetask(ctx, tm):
    start_time = time.time()
    await asyncio.sleep(tm)
    return time.time() - start_time
```
it works as expected, raising `asyncio.exceptions.TimeoutError`.

I can see the timeout is being managed using `async-timeout`. Perhaps some changes are needed to deal with job functions that rarely yield to the asyncio loop.

cmangla:
It seems async-timeout authors actually intend the semantics leading to my issue, as can be seen in their test here:

```Python3
@pytest.mark.asyncio
async def test_timeout_blocking_loop() -> None:
    async def long_running_task() -> str:
        time.sleep(0.1)
        return "done"


    async with timeout(0.01):
        result = await long_running_task()
    assert result == "done"
```
https://github.com/aio-libs/async-timeout/blob/7b9302be7f00bf6fbb21cf2592a35061ca4bdabe/tests/test_timeout.py#L88-L96

Therefore, in arq, it might be worth checking if the timeout deadline was long expired and noting it as a timeout.

cmangla:
As explained by @samuelcolvin and @Kludex in #227, blocking tasks should be run in an executor, which can be applied to this example like so:
```Python3
import time
import asyncio
from arq import create_pool
from arq.connections import RedisSettings
from concurrent import futures
import functools

async def thetask(ctx, tm):
    sleepy = functools.partial(time.sleep, tm) #(*1)
    loop = asyncio.get_running_loop()
    start_time = time.time()
[...]

PR #232, opened 2021-02-28
Add example for sync jobs to docs

cmangla:
See issue #230

### DIFF
--- a/docs/examples/sync_job.py
+++ b/docs/examples/sync_job.py
@@ -1,3 +1,8 @@
+import time
+import functools
+import asyncio
+import futures
+
 def sync_task(t):
     return time.sleep(t)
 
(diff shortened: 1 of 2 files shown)


docs/examples/sync_job.py
```diff
@@ -0,0 +1,14 @@
+def sync_task(t):
```
samuelcolvin, line 1: please include all imports required to run the example.

samuelcolvin (commented):
otherwise looks good to me.

docs/examples/sync_job.py
cmangla: Done.

samuelcolvin:
Thanks so much.

Issue #238, opened 2021-04-19
Cron jobs fail with msgpack serializaiton


Artur-Biniek:
Minimal reproducible example:

``` python
from arq import cron
import msgpack

async def run_regularly(ctx, *args, **kwargs):
    print('run every 15 secs\n')

class WorkerSettings:
    cron_jobs = [cron(run_regularly, second=(0, 15, 30, 45))]
    job_serializer = msgpack.packb
    job_deserializer = lambda b: msgpack.unpackb(b, raw=False)
```

Worker fails with error similar to:
```
   job_deserializer = lambda b: msgpack.unpackb(b, raw=False)
  File "msgpack/_unpacker.pyx", line 202, in msgpack._cmsgpack.unpackb
msgpack.exceptions.ExtraData: unpack(b) received extra data.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File ".../.venv/lib/python3.9/site-packages/arq/worker.py", line 395, in run_job
    function_name, args, kwargs, enqueue_job_try, enqueue_time_ms = deserialize_job_raw(
  File ".../.venv/lib/python3.9/site-packages/arq/jobs.py", line 231, in deserialize_job_raw
    raise DeserializationError('unable to deserialize job') from e
arq.jobs.DeserializationError: unable to deserialize job
16:03:48: error serializing result of cron:run_regularly:1618841025123:<unknown>
```

kornicameister:
@samuelcolvin  @Artur-Biniek problem is not about `msgpack`, it is about how `arq` initializes a redis pool. If you look at [here](https://github.com/samuelcolvin/arq/blob/master/arq/worker.py#L278, you will see that if pool is not set `arq` attempts to create it knowing redis pool settings.
What's not passed "downstream" is the remaining piece of information concerning serialization functions in particular.

kornicameister:
@samuelcolvin I've added a small PR that helps out here. Not sure if that's a correct fix. Perhaps a better approach, at least for future, would be to either allow setting just the pool or just the settings + have the latter extended with details about `job_serialization`, for instance.

samuelcolvin:
I think this is fixed my #203

PR #218, opened 2020-12-17
create worker pool with worker's queue

erakli:

PR #239, opened 2021-04-20
Let pool know about other worker's settings

kornicameister:
Allows `ArqRedis` instance, being a member of `Worker`, know about `default_queue_name` and serialization mechanism.

---

Fixes: #238


samuelcolvin:
thank you.

I believe this is already fixed by by #203 and #218

kornicameister:
indeed it is! cheers!

PR #258, opened 2021-07-30
fix: arq is incompatible for aioredis>=2.0.0

PaxPrz:
arq is currently breaking because, it now installs aioredis 2.0.0 and this version is not compatible as errors are now changed to exceptions and so on...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ setup(
         arq=arq.cli:cli
     """,
     install_requires=[
-        'aioredis>=1.1.0',
+        'aioredis>=1.1.0,<2.0.0',
         'click>=6.7',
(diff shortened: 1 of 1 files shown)


JrooTJunior:
The best way to fix compatibility with `aioredis` 2.0, as it is already released.

In this case, it makes no sense to restrict version and publish new release of `arq` without compatibility with 2.0

teror4uks:
> The best way to fix compatibility with aioredis 2.0, as it is already released.

Fixing compatibility could spend more time than just release the package with a restriction. It solves the problem in many pipelines without additional work from their side.

teror4uks:
@samuelcolvin could you please have a look at this?

Issue #250, opened 2021-05-25
AioRedis v2 Changes / Migration


jtbaker:
A project I'm working on needs to use both `arq` and `aioredis` to connect to different redis databases with separate use cases. Following the `aioredis` docs for using their 2.x release as instructed here https://github.com/aio-libs/aioredis-py/issues/987, I installed using the 2.0.0a1 tag. After installing this dependency, running `import arq` throws this stack trace:

```
config.py:2: in <module>
    import arq
../../../venv/lib/python3.9/site-packages/arq/__init__.py:1: in <module>
    from .connections import ArqRedis, create_pool  # noqa F401
../../../venv/lib/python3.9/site-packages/arq/connections.py:13: in <module>
    from aioredis import MultiExecError, Redis
E   ImportError: cannot import name 'MultiExecError' from 'aioredis'
```

I guess with the 2.0 `aioredis` release, the naming/location of the module's exceptions, and perhaps other module internals have changed. Is migrating to to use this recent aioredis release a goal of the project? I'd be happy to dig in to it further and possibly put together a PR if it was deemed appropriate.

Thanks for your work on this project, it's been quite useful.

SamiAlsubhi:
aioredis 2.0 is not finalized and not stable yet.

dannya:
Seems that aioredis 2.0 has just been finalized with a stable release:

https://github.com/aio-libs/aioredis-py/releases

havardthom:
+1

I'm unable to upgrade aioredis to 2.0.0 in my project due to arq still being on 1.x version

JrooTJunior:
@samuelcolvin, do you have any plans to fix compatibility with aioredis 2.0?

samuelcolvin:
I'm keen, but I'm super busy atm. I'll look soon.

waza-ari:
Honestly @Olegt0rr you have a strange attitude towards open source projects maintained on a private and voluntary basis. 

@samuelcolvin I feel this need to be said: thanks for all your work, and congratulations to your new family member. Enjoy the time as much as you can!

The only thing I'm concerned with is that the library as it is right now is broken for fresh installs, as `arq/setup.py` contains:

```
install_requires=[
 'aioredis>=1.1.0',
```

So by default, it will now install aioredis 2.0.0, which is incompatible with arq. Short-term option is to manually install `aioredis` and force it to use a 1.x version:

```
pipenv install aioredis=='1.3.1'
```

[...]

PR #259, opened 2021-08-07
support aioredis v2

Yolley:
This is my attempt at aioredis2 support, I've made these changes while aioredis was in beta, but they are 100% compatible with release version. All tests pass, but I needed to change some of them, there was an issue with pipe, I did not manage to reproduce multi-exec error, so I commented 2 asserts in `test_multi_exec `. Also because there is no way to decode response per call in aioredis2, I added `.decode()` in some places, I am not happy with this solution, may be we need to have 2 separate clients: one for pickled objects fetch, one for everything else, I am not sure. Also with this pr we will totally drop support for aioredis v1. Overall I was making this feature in a bit of a rush, therefore it will be a draft pr, may be it can become a base for the actual pr/release.


samuelcolvin:
Thanks so much for working on this @Yolley. I really want to get it merged and released.

Do you want  to work on it and great it ready for full review? Then I'll review and we can go from there.

I agree the decoding situation is a mess, I assume we'd have no luck asking the maintainers of aioredis to make decoding a per-command option since they want to stick to the redis-py interface?

I don't want to have two connections/pools just for this. I think we should implement our own `get` command with an `encoding` argument, but I haven't played with aioredis 2 enough yet to really be sure.

Yolley:
> Thanks so much for working on this @Yolley. I really want to get it merged and released.
> 
> Do you want to work on it and great it ready for full review? Then I'll review and we can go from there.
> 
> I agree the decoding situation is a mess, I assume we'd have no luck asking the maintainers of aioredis to make decoding a per-command option since they want to stick to the redis-py interface?
> 
> I don't want to have two connections/pools just for this. I think we should implement our own `get` command with an `encoding` argument, but I haven't played with aioredis 2 enough yet to really be sure.

I want to work on it, but I am not sure when I get the time to polish this PR. I agree with implementing our own `get` method in ArqRedis (at least until aioredis-py supports it in some way, if they agree to do so).

Yolley:
Okay, it seems that it won't be that easy, they have 2 different parsers that can decode responses: `PythonParser` that uses `Encoder` https://github.com/aio-libs/aioredis-py/blob/master/aioredis/connection.py#L342, `Encoder` has `force` flag https://github.com/aio-libs/aioredis-py/blob/master/aioredis/connection.py#L131 that we can try to utilize, but `read_response ` in `PythonParser` (it decodes the response) does not accept any parameters, so we will need to write a subclass of this parser it seems.
And there is also `HiredisParser` https://github.com/aio-libs/aioredis-py/blob/master/aioredis/connection.py#L412, it uses encoding from `hiredis`, so it seems that we will also need to write our wrapper for `hiredis` responses, that will actually decode the response (that's how they did it in aioredis-py v1 it seems https://github.com/aio-libs/aioredis-py/blob/62a75ba8cac8236eff6836248e1be7b68af1b37a/aioredis/parser.py#L103 )

sondrelg:
Thanks for working on this @Yolley 👏 

Is there anything others can do to help make progress on this? 🙂

Yolley:
@samuelcolvin so, I've decided to go with context manager and a context variable, so that we could decode on per-request basis (I decided to stick with context vars to not subclass almost everything from aioredis-py). This change will drop support for python 3.6 (because context vars became introduced in 3.7), also I've turned off type checks for `parser.py`, because it contains subclasses of aioredis parser, and currently aioredis is poorly annotated. All tests pass, but I guess I will need to add some more tests (may be for hiredis and for context encoding itself), but first please take a look at my changes and tell, if you are okay with this approach.

Yolley:
@samuelcolvin I've marked this PR ready for merge, please take a look when you get the chance, I at least need to understand if my view of aioredis 2 support in arq coincides with yours.

arq/connections.py
```diff
@@ -129,14 +143,15 @@ async def enqueue_job(
...
-            pipe.watch(job_key)
-            job_exists = pipe.exists(job_key)
-            job_result_exists = pipe.exists(result_key_prefix + job_id)
-            await pipe.execute()
-            if await job_exists or await job_result_exists:
+            await pipe.unwatch()
+            await pipe.watch(job_key)
+            job_exists = await pipe.exists(job_key)
+            job_result_exists = await pipe.exists(result_key_prefix + job_id)
+            # job_exists, job_result_exists = await pipe.execute()
```
JonasKs, line 152: Left over comment

arq/connections.py
```diff
@@ -150,19 +165,19 @@ async def enqueue_job(
...
+            pipe.zadd(_queue_name, {job_id: score})
             try:
-                await tr.execute()
-            except MultiExecError:
+                await pipe.execute()
+            except (ResponseError, WatchError):
                 # job got enqueued since we checked 'job_exists'
                 # https://github.com/samuelcolvin/arq/issues/131, avoid warnings in log
-                await asyncio.gather(*tr._results, return_exceptions=True)
+                # await asyncio.gather(*tr._results, return_exceptions=True)
```
JonasKs, line 176: Left over comment

arq/worker.py
```diff
@@ -378,29 +380,29 @@ async def start_jobs(self, job_ids: List[str]) -> None:
...
-                    await tr.execute()
-                except MultiExecError:
+                    await pipe.execute()
+                except (ResponseError, WatchError):
                     # job already started elsewhere since we got 'existing'
                     self.sem.release()
                     logger.debug('multi-exec error, job %s already started elsewhere', job_id)
                     # https://github.com/samuelcolvin/arq/issues/131, avoid warnings in log
-                    await asyncio.gather(*tr._results, return_exceptions=True)
+                    # await asyncio.gather(*tr._results, return_exceptions=True)
```
JonasKs, line 405: Left over comment

JonasKs (commented):
Not a review of the code, but seems like you've forgot to remove a few comments.

JonasKs:
Just letting you know that I've implemented your fork in a project, and everything works very well so far!

elrik75:
I tried this PR and I have this issue: when I restart Redis/Sentinel (while the worker is still running) I have this error:
```
Traceback (most recent call last):
  File "./catwalk/redis/rpc.py", line 105, in launch_worker
    await worker.main()
  File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/arq/worker.py", line 308, in main
    await self._poll_iteration()
  File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/arq/worker.py", line 347, in _poll_iteration
    t.result()
  File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/arq/worker.py", line 419, in run_job
    v, job_try, _ = await asyncio.gather(*coros)
  File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/aioredis/client.py", line 1064, in execute_command
    return await self.parse_response(conn, command_name, **options)
  File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/aioredis/client.py", line 1080, in parse_response
    response = await connection.read_response()
  File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/aioredis/sentinel.py", line 65, in read_response
    return await super().read_response()
  File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/aioredis/connection.py", line 854, in read_response
    response = await self._parser.read_response()
  File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/arq/parser.py", line 110, in read_response
    response = self._reader.gets()
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```

It tries to read the content as UTF-8 but it's a binary pickle data.

Yolley:
> I tried this PR and I have this issue: when I restart Redis/Sentinel (while the worker is still running) I have this error:
> 
> ```
> Traceback (most recent call last):
>   File "./catwalk/redis/rpc.py", line 105, in launch_worker
>     await worker.main()
>   File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/arq/worker.py", line 308, in main
>     await self._poll_iteration()
>   File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/arq/worker.py", line 347, in _poll_iteration
>     t.result()
>   File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/arq/worker.py", line 419, in run_job
>     v, job_try, _ = await asyncio.gather(*coros)
>   File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/aioredis/client.py", line 1064, in execute_command
>     return await self.parse_response(conn, command_name, **options)
>   File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/aioredis/client.py", line 1080, in parse_response
>     response = await connection.read_response()
>   File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/aioredis/sentinel.py", line 65, in read_response
>     return await super().read_response()
>   File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/aioredis/connection.py", line 854, in read_response
>     response = await self._parser.read_response()
>   File "/home/elrik/Info/SecretSauce/cat-walk/venv/lib/python3.7/site-packages/arq/parser.py", line 110, in read_response
>     response = self._reader.gets()
> UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
> ```
> 
> It tries to read the content as UTF-8 but it's a binary pickle data.
> 
> edit: I've found a way to go around this issue by fully re-creating the worker when Redis is interrupted.

Does it only happen with Sentinel, did you restart every instance or only main? I've tried to restart a Redis (without Sentinel) instance, the worker just failed
```aioredis.exceptions.ConnectionError: Error connecting to localhost:6379. Multiple exceptions: [Errno 61] Connect call failed ('::1', 6379, 0, 0), [Errno 61] Connect call failed ('127.0.0.1', 6379).
```
There is no handlers for these errors in current arq release, so it's not a regression.

elrik75:
Yes I use Sentinel (3 redis of which 2 replicas, 3 sentinels)
Basically I cannot stop my process because redis/sentinel fails, I have 3 instances and Sentinel will do its job and I will have a new Redis Master soon. So I have a system that tries to reconnect to redis every second. It worked well until now:
with Arq, I used a task that relaunch `worker.main()` after a redis error. But I got the issue. As @samuelcolvin  wrote about encoding issue with `aioredis2`, I thought it was maybe an issue of this PR. In fact, I did not try with `aioredis1.3 `as all my Redis code will fail (I use a lot of Redis, not only Arq).
As I said, I found a way that works well by re-creating all the workers after a Redis disconnection.

arq/worker.py
```diff
@@ -409,7 +409,7 @@ async def start_jobs(self, job_ids: List[str]) -> None:
     async def run_job(self, job_id: str, score: int) -> None:  # noqa: C901
         start_ms = timestamp_ms()
         coros = (
-            self.pool.get(job_key_prefix + job_id, encoding=None),
+            self.pool.get(job_key_prefix + job_id),
```
samuelcolvin, line 412: we should use a pipeline here I think.

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [tool:pytest]
 testpaths = tests
-timeout = 5
+;timeout = 5
```
samuelcolvin, line 3: good to keep something here, are some tests really taking more than 5 seconds?

arq/parser.py
```diff
@@ -0,0 +1,126 @@
+#  type: ignore
```
samuelcolvin, line 1: This is my main question about this PR (and about migrating to aioredis v2 in general).

In general, it's a real shame that the developers of aioredis 2 decided to constraint the encode/decode decision to be made on the connection/pool, not on an individual command. I assume this was so they code do (or not do) the encoding in hiredis and improve performance?

What you've written here is very cool and a sophisticated approach, but it doesn't actually avoid backwards compatibility problems, and in some ways makes the logic within the library more complex. I also wonder if it will incur a performance penalty?

It makes me sad to say it, but I wonder if we should removing this and just use `.decode()` everywhere that we want a string?

Always returning bytes and forcing the user/developer to call `.decode()` is definitely the explicit keep-it-simple-stupid approach.

For the tests we can just compare to bytes, no need for either `with self.pool.encoder_context...` or `.decode()`.

It will mean in application logic that users (including me!) have to add `.decode()` in a lot of places, but the alternative is to add `with self.pool.encoder_context...` in nearly as many places - there's no route that doesn't cause a big fat change to the API.

arq/connections.py
```diff
@@ -129,14 +143,14 @@ async def enqueue_job(
...
-        with await self as conn:
+        async with self as conn:
             pipe = conn.pipeline()
-            pipe.unwatch()
-            pipe.watch(job_key)
-            job_exists = pipe.exists(job_key)
-            job_result_exists = pipe.exists(result_key_prefix + job_id)
-            await pipe.execute()
-            if await job_exists or await job_result_exists:
+            await pipe.unwatch()
```
samuelcolvin, line 148: Maybe you can remove `unwatch`? It was required since with aioredis<2 connections often still had a watch in place from previous commands

arq/connections.py
```diff
@@ -150,19 +164,17 @@ async def enqueue_job(
...
-            tr.psetex(job_key, expires_ms, job)
-            tr.zadd(_queue_name, score, job_id)
+            pipe.multi()
+            pipe.psetex(job_key, expires_ms, job)
+            pipe.zadd(_queue_name, {job_id: score})
             try:
-                await tr.execute()
-            except MultiExecError:
+                await pipe.execute()
+            except (ResponseError, WatchError):
```
samuelcolvin, line 172: Not sure we need to catch `ResponseError` here too? This case is specifically for when the `watched` key get's set.

samuelcolvin (commented):
overall this looks great, a few fairly small things to change and one big decision on `parser.py`

PerAStad:
Hi all!

I seem to having some issues using the `arq.Retry` raising method ([link](https://arq-docs.helpmanual.io/#retrying-jobs-and-cancellation)) for scheduled cron jobs.

## Use Case
- Update a token with a scheduled task from a third party application, ensuring my backend to always have a valid token at hand. 
- The third party application tends to respond with a gateway-timeout-error or something similar. Reason behind this can be many things, and can often times be fixed by simply performing another equal request, after a few seconds. 
  - I therefore wish to implement a sort of `cron job failed --> retrying job after X seconds` policy. 
  - By the looks of it, this seems to be a valid use case for the `arq.Retry` method. 

This is my `arq.py` file:
```python
import logging

from arq.cron import cron
from arq import Retry

from app.core.config import settings
from app.core.logging_config import setup_logging
import asyncio

log = logging.getLogger(__name__)


async def create_token(ctx: dict) -> None:
    """
    Create token and saving it to Redis
    """
    log.info('Creating token')
    # Pretending we fetching a token here:
    for attempt in range(2):
        if attempt != 1:
            # Take a little nap
            await asyncio.sleep(5)
            raise Retry(defer=ctx['job_try'] * 5)

    log.info('Token has been successfully created.')


async def startup(ctx: dict) -> None:
    """
    On startup. 
    """
    setup_logging()
    log.info('Arq worker is starting...')
    # await create_token(ctx)
    log.info('Arq worker startup was successful.')


class WorkerSettings:
    """
    ARQ Worker settings.
    """

    update_token_cron_job = cron(create_token, second={0, 30})

    cron_jobs = [update_token_cron_job]
    on_startup = startup

    redis_settings = settings.REDIS_SETTINGS

```
Running: `arq app.core.arq.WorkerSettings --watch app`:

Starts the arq worker correctly until the actual `update_token_cron_job` job is executed. This is the terminal output:
image

With the following error:
```
  File "/home/user/project/.venv/lib/python3.9/site-packages/aioredis/connection.py", line 868, in read_response
    raise response from None
aioredis.exceptions.ResponseError: Command # 3 (DEL) of pipeline caused error: ("wrong number of arguments for 'del' command",)
```

Also, if I include the job in the startup (uncommenting `await create_token(ctx)` under startup) sequence there is a keyError regarding `'job_try'`:
```
  File "/home/user/project/app/core/arq.py", line 23, in create_token
    raise Retry(defer=ctx['job_try'] * 5)
KeyError: 'job_try'
```

Examining the [docs](https://arq-docs.helpmanual.io/#retrying-jobs-and-cancellation), it can be seen that they are not using the `Retry` method for cron jobs but rather functions. Have I misunderstood how to use this `Retry` method, or is this a bug? All inputs are welcome!

Thank you for taking the time reading this :)

samuelcolvin:
@PerAStad please can you move this to a new issue, optionally referencing this PR.

elrik75:
Just a note to tell that I've used Arq + aioredis2 for 2 months now in production and it works well.

arq/connections.py
Yolley: I am pretty sure that at some point in my tests I got this error, now I can't reproduce it, removed it from except block.

arq/connections.py
Yolley: Okay, ran tests several times, seems to be working fine without unwatch.

setup.cfg
Yolley: Oh, I overlooked it, I added it when I was working on the changes for v2, needed to run tests without timeout. Currently in my environment only `test_redis_sentinel_failure` fails because of timeout, and I think it should be expected, because with 5 retries and 1 second sleep `test_redis_sentinel_failure` will always run for at least 5 seconds. I think we need to fix this test (may be decrease max retries?).

arq/worker.py
Yolley: You are right, added the pipeline.

arq/parser.py
```diff
@@ -0,0 +1,126 @@
...
+
+    HIREDIS_AVAILABLE = True
+except (ImportError, ModuleNotFoundError):
+    HIREDIS_AVAILABLE = False
+
+
+encoder_options_var: ContextVar[Optional[Dict[str, Any]]] = ContextVar('encoder_kwargs', default=None)
+
+
+class ContextAwareEncoder(Encoder):
```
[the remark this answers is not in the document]
Yolley: Added some description to the classes.

arq/connections.py
```diff
@@ -129,14 +143,14 @@ async def enqueue_job(
...
-            job_exists = pipe.exists(job_key)
-            job_result_exists = pipe.exists(result_key_prefix + job_id)
-            await pipe.execute()
-            if await job_exists or await job_result_exists:
+            await pipe.unwatch()
+            await pipe.watch(job_key)
+            job_exists = await pipe.exists(job_key)
+            job_result_exists = await pipe.exists(result_key_prefix + job_id)
+            if job_exists or job_result_exists:
+                await pipe.reset()
```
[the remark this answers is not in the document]
Yolley: Not sure what exactly you wanted to chain here, unwatch and watch should be executed one after another, so I chained just 2 `exists` in a `gather`.

arq/parser.py
Yolley: I think their reasoning was just to port redis-py. Because redis-py does not support decode on per-request basis, they also decided not to add it. But redis-py maintainer actually thought of adding this feature, they even add to hiredis an ability to change encoding settings at runtime https://github.com/redis/hiredis-py/pull/96 (and we use it in `ContextAwareHiredisParser`). Also I am pretty sure that I saw a branch in `redis-py` repository where they started working on the encoding context, but it seems it was removed at some point (so may be they decided against it).

In terms of performance, as I understand, changing encoding on HiRedis should not be a huge overhead in terms of performance, this method just updates parameters. But I am not sure about how much context variables handling can influence performance. I can run some tests later.

But if you really think that we should stick with simpler approcah, I can remove context encoding and return to `decode` calls (basically what I did with first set of changes).

Yolley:
I've added a conversion for expire parameters to milliseconds, because aioredis-py v2 requires EX or PX to be provided as integer, so I just decided to convert floats to milliseconds and pass it as PX to aioredis calls. Also added a test for it (will fail without these changes).

arq/parser.py
Yolley: They even mention encoding context manager in `lock.py` https://github.com/aio-libs/aioredis-py/blob/master/aioredis/lock.py#L235 (I guess it was just copied from `redis-py`).

samuelcolvin:
thank you so much @Yolley for all your work on this. Since it's now passing I've merge this PR, although there's a bit more work to do before a release. I'll comment on #250.

PR #261, opened 2021-08-16
fix: package importing in example

cdpath:


samuelcolvin:
good catch, thanks so much.

PR #273, opened 2021-10-30
fix error when cron jobs were terminanted

### LINKED ISSUE #253: Cron job termination error
Hi, 
when worker is terminated with CTRL+C while any cron job is running there is exception: 
`aioredis.errors.MultiExecError: ('MultiExecError errors:', [TypeError("delete() missing 1 required positional argument: 'key'")])`

It happens here: https://github.com/samuelcolvin/arq/blob/8ee04f96934684c9b8f8278df0530ae9fad1be4a/arq/worker.py#L625 because `delete_keys` is empty. Obvious fix would be to check if `delete_keys` is not empty before doing `tr.delete(*delete_keys) ` but I'm not sure if it's expected situation or problem is deeper.

tobymao:
fixes #253

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -620,7 +620,8 @@ class Worker:
                 tr.zrem(self.queue_name, job_id)
             elif incr_score:
                 tr.zincrby(self.queue_name, incr_score, job_id)
-            tr.delete(*delete_keys)
+            if delete_keys:
+                tr.delete(*delete_keys)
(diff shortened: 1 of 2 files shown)


tobymao:
@samuelcolvin hello, any chance these PR's will get reviewed?

samuelcolvin:
Thanks for this, it's been annoying me too.

PR #274, opened 2021-10-30
add on_job_start and on_job_end hooks

tobymao:


arq/worker.py
```diff
@@ -509,6 +515,10 @@ async def job_failed(exc: BaseException) -> None:
             'score': score,
         }
         ctx = {**self.ctx, **job_ctx}
+
+        if self.on_job_start:
```
aovasylenko, line 519: should it be wrapped in try .. except?

tobymao: i chose not to do this because on_startup and on_shutdown also don't wrap a try catch, if you think it is needed, i can add it.

it certainly adds a lot more complexity though because you need to check all the different types of exceptions

samuelcolvin:
great, thank you.

AngellusMortis:
This is a really great new feature I could use in the project I am writing now trying to use arq instead of celery. 

Is there anything I can do to help get released to a pypi version?

PR #275, opened 2021-10-30
allow storage of traceback on exception

tobymao:


samuelcolvin:
I think better to implement this in your own logic. In general, something has gone wrong when you're inspecting the string representation of a worker error in the client.

PR #278, opened 2021-11-08
Update argument docstring definition

sondrelg:
Looks like the `queue_read_limit` argument was described correctly when initially introduced in https://github.com/samuelcolvin/arq/pull/142, but that the default value was later changed to fix an issue (in https://github.com/samuelcolvin/arq/pull/180) without the docstring being updated.


samuelcolvin:
thank you.

PR #279, opened 2021-11-09
Add link to WorkerSettings in documentation

JonasKs:
A minor detail I'd like to add to the docs. Obviously not necessary, but it took me longer than I'd like to admit to troubleshoot why the worker would not start (I had connection startup configured in my `startup` function instead) :smile:

### DIFF
--- a/docs/examples/main_demo.py
+++ b/docs/examples/main_demo.py
@@ -22,7 +22,8 @@ async def main():
         await redis.enqueue_job('download_content', url)
 
 # WorkerSettings defines the settings to use when creating the work,
-# it's used by the arq cli
+# it's used by the arq cli.
+# For a list of available settings, see https://arq-docs.helpmanual.io/#arq.worker.Worker
 class WorkerSettings:
     functions = [download_content]
     on_startup = startup


samuelcolvin:
this isn't necessary in most contexts, only if you have a custom `RedisSettings`. Better to add a comment to the code saying "include `redis_settings = RedisSettings(...)` here if you want to customise the connection?

JonasKs:
Sure! I'll fix today.

samuelcolvin:
thanks so much.

RamonGiovane:
I faced this issue more than  once and I think the docs are not clear enough yet.

It should explain that if you're not using the default settings (which is the most common case for production uses)  you must config `RedisSettings` twice: 

At the WorkerSettings class, statically
```python
class WorkerSettings:
    functions = [download_content]
    on_startup = startup
    on_shutdown = shutdown
    redis_settings=RedisSettings() # HERE !!
```

and when instantiating the pool to enqueue a job:
```python
async def main():
    redis = await create_pool(RedisSettings()) # HERE !!
    for url in ('https://facebook.com', 'https://microsoft.com', 'https://github.com'):
        await redis.enqueue_job('download_content', url)
```        

This is the problem reported in #361 and #305.
`WorkerSettings` assumes a `redis_settings` **implicitly**, but users usually think that passing this settings once at the `create_pool` call is enough when it isn't.

PR #281, opened 2021-12-01
the enqueue_time in context can use the direct timezone

ponytailer:


ponytailer:
@samuelcolvin please

ponytailer:
> I think I would rather stick to all timestamps being UTC, I don't want to add more complexity and depdencies unless absolutely necessary.
> 
> 
> 
> Please can you provide an explanation of why you need this and how you will use it beyond "it's not good.".

In my country，the default timezone is not the utc。So when the job is processing，the log was less 8 hours。 Not easy to read for people who is not the programmer。

samuelcolvin:
In my country, the default timezone is not UTC either. :-) Arguably it's even worse, because the timezone IS utc for half the year, then UTC+1 for the other half.

Still, as far as I know arq only shows time differences in logs. If you want to log datetimes, in your own timezone you can implement that in your own code without modifying arq.

I'm inclined to close this, unless there's some piece of code I'm missing that lots datetimes and would require a change to the code base to log correctly.

ponytailer:
> In my country, the default timezone is not UTC either. :-) Arguably it's even worse, because the timezone IS utc for half the year, then UTC+1 for the other half.
> 
> 
> 
> Still, as far as I know arq only shows time differences in logs. If you want to log datetimes, in your own timezone you can implement that in your own code without modifying arq.
> 
> 
> 
> I'm inclined to close this, unless there's some piece of code I'm missing that lots datetimes and would require a change to the code base to log correctly.

but when the job was finished or others status。arq will print the log include the utc datetime，it's not good for reading。
unless write my own arq worker class extends Worker and rewrite it。 it's complex.

arq/constants.py
```diff
@@ -10,3 +10,11 @@
 # how long to keep the "in_progress" key after a cron job ends to prevent the job duplication
 # this can be a long time since each cron job has an ID that is unique for the intended execution time
 keep_cronjob_progress = 60
+
+arq_timezone = "ARQ_TIMEZONE"
```
samuelcolvin, line 14: single quotes please.

arq/utils.py
```diff
@@ -26,7 +30,23 @@ def to_unix_ms(dt: datetime) -> int:
 
 
 def ms_to_datetime(unix_ms: int) -> datetime:
-    return datetime.fromtimestamp(unix_ms / 1000, tz=timezone.utc)
+    """
+    convert milliseconds to datetime, use the timezone in os.environ
+    """
+    tz = None
+    for timezone_key in timezone_keys:
```
samuelcolvin, line 37: this would have a massive impact on performance as we would be checking multiple env keys every time we called this function.

Please move the logic of getting the tz out into a new function, then use `functools.cache` to avoid calling this logic multiple times.

arq/utils.py
```diff
@@ -26,7 +30,23 @@ def to_unix_ms(dt: datetime) -> int:
...
+        tz = os.environ.get(timezone_key)
+        if tz:
+            break
+
+    if tz is not None and isinstance(tz, str):
```
samuelcolvin, line 42: these two checks are redundant. `if tz:` is sufficient.

arq/utils.py
```diff
@@ -26,7 +30,23 @@ def to_unix_ms(dt: datetime) -> int:
...
+        try:
+            tz = pytz.timzone(tz)
+        except UnknownTimeZoneError:
+            tz = timezone.utc
```
samuelcolvin, line 46: this error should be raised, or at least warned about. Silently catching errors like this is very confusing.

arq/utils.py
```diff
@@ -1,8 +1,12 @@
 import asyncio
 import logging
+import os
+import pytz
```
samuelcolvin, line 4: pytz is not a dependency of this package, and we're not adding it.

You need to catch `ImportError` and only check for other timezones if it's installed.

arq/utils.py
```diff
@@ -1,8 +1,12 @@
...
+from pytz.exceptions import UnknownTimeZoneError
```
samuelcolvin, line 5: you don't need to use this specific error, it inherits from `KeyError`.

arq/worker.py
```diff
@@ -247,6 +251,9 @@ def __init__(
         self.job_serializer = job_serializer
         self.job_deserializer = job_deserializer
 
+        if timezone is not None and isinstance(timezone, str):
+            os.environ[arq_timezone] = timezone
```
samuelcolvin, line 255: why are you doing this? This is not a good way to shared variables globally.

Either use an actual global variable, or use an actual environment variable.

This is an antipattern.

setup.py
```diff
@@ -58,6 +58,7 @@
         'aioredis>=1.1.0,<2.0.0',
         'click>=6.7',
         'pydantic>=1',
+        'pytz',
```
samuelcolvin, line 61: No need to add pytz as a dependency, post people won't use it.

And if you did make it a dependency, you'd need to restrict it.

samuelcolvin (commented):
Okay, I'm prepared to accept this change, but this pull request needs a lot of work.

ponytailer:
> Okay, I'm prepared to accept this change, but this pull request needs a lot of work.

Thanks for your review.

samuelcolvin:
great, thanks so much.

PR #293, opened 2022-02-17
Allow setting job_id on cron jobs

### LINKED ISSUE #292: Cron job uniqueness/singleton
Hi, 

I can see that the `cron` jobs have a `unique=True` flag, which allows us to run multiple workers, without spawning the same cron job multiple times. This parameter does not ensure the `uniqueness` described in the [documentation](https://arq-docs.helpmanual.io/#job-uniqueness):

> Sometimes you want a job to only be run once at a time (eg. a backup) or once for a given parameter (eg. generating invoices for a particular company).

but instead that even if you have 10 workers, the task will only be run once. Which is awesome! But I need both. :smile: 

------ 

Let's say we have backup jobs running every 10 minutes at 12:00, 12:10, 12:20, 12:30 etc. Normally this job is done in 5 minutes, but this time the backup job at 12:20 took 15 minutes, so we want to skip the 12:30 job, and wait for the 12:40 job. Something like this:

```
cron schedule
  ^   |
12:10 |          xxxxx
12:20 |                      xxxxxxxxxxxxxxxxxx
12:30 |                                  SKIP
12:40 |                                               xxxxx
       ----------10----------20----------30----------40----------50----
                            time in minutes
```

This pattern could be achieved by allowing specifying the `job_id` on cron jobs too. I'll submit a PR, and if you accept the solution I will add tests.

JonasKs:
Closes #292 

I will add tests if you accept the feature.

### DIFF
--- a/arq/cron.py
+++ b/arq/cron.py
@@ -102,6 +102,7 @@ class CronJob:
     microsecond: int
     run_at_startup: bool
     unique: bool
+    job_id: Optional[str]
     timeout_s: Optional[float]
     keep_result_s: Optional[float]
     keep_result_forever: Optional[bool]
@@ -137,6 +138,7 @@ def cron(
     microsecond: int = 123_456,
     run_at_startup: bool = False,
     unique: bool = True,
+    job_id: Optional[str] = None,
     timeout: Optional[SecondsTimedelta] = None,
     keep_result: Optional[float] = 0,
     keep_result_forever: Optional[bool] = False,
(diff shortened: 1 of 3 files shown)


arq/cron.py
```diff
@@ -159,7 +161,8 @@ def cron(
     :param microsecond: microsecond(s) to run the job on,
         defaults to 123456 as the world is busier at the top of a second, 0 - 1e6
     :param run_at_startup: whether to run as worker starts
-    :param unique: whether the job should be only be executed once at each time
+    :param unique: whether the job should be only be executed once at each time (useful if you have multiple workers)
```
sondrelg, line 164: Think there's a "be" in here too many

JonasKs: Fixed. Thanks :blush:

samuelcolvin:
I'm happy with this, please rebase to master and add tests.

JonasKs:
Rebased and added a test, @samuelcolvin.

Not entirely sure how we should test this without using burst, so I used burst + two different functions. Two cron jobs using the same function with `run_at_startup=True` would still only be run once.

samuelcolvin:
great, thank you.

PR #294, opened 2022-02-17
Add --custom-log-dict CLI option

isobelhooper:
This points at an import path for a dictionary (e.g. myproject.logging.CONFIG_DICT) that Arq can configure its logging with instead of using the default configuration.

This would be useful for e.g. configuring Arq logging output to use [python-json-logger](https://github.com/madzak/python-json-logger).

We use this in a fork of Arq for a project I'm working on, and I wanted to contribute this back upstream. I'm happy to make any changes that are needed.

### DIFF
--- a/arq/cli.py
+++ b/arq/cli.py
@@ -1,4 +1,5 @@
 import asyncio
+import importlib
 import logging.config
 import os
 import sys
@@ -19,6 +20,7 @@ burst_help = 'Batch mode: exit once no jobs are found in any queue.'
 health_check_help = 'Health Check: run a health check and exit.'
 watch_help = 'Watch a directory and reload the worker upon changes.'
 verbose_help = 'Enable verbose output.'
+logdict_help = "Import path for a dictionary in logdict form, to configure Arq's own logging."
 
 
 @click.command('arq')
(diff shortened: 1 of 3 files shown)


arq/cli.py
```diff
@@ -28,15 +30,24 @@
...
     """
     sys.path.append(os.getcwd())
     worker_settings_ = cast('WorkerSettingsType', import_string(worker_settings))
-    logging.config.dictConfig(default_log_config(verbose))
+    if custom_log_dict:
+        try:
+            config_path, config_dict = custom_log_dict.rsplit('.', maxsplit=1)
+            log_config = getattr(importlib.import_module(config_path), config_dict)
+        except (TypeError, AttributeError):
+            log_config = default_log_config(verbose)
```
samuelcolvin, line 47: we should log (probably just print to sterr is easiest) a warning in this case with details of what went wrong.

Or we could just raise an error.

samuelcolvin (commented):
I'm okay with this, needs also:
* a test
* to be rebased to master - should fix build

arq/cli.py
isobelhooper: Raising an ImportError from `import_string` seems like it would work okay - I'll do that.

samuelcolvin:
you've got an error in CI on py3.10. You (or I) will also need to pin jinja to make docs build.

samuelcolvin:
update, I've pinned jinja in master.

isobelhooper:
This is now rebased on current master, and I've also updated the version of black to deal with an incompatibility with certain Click versions. Let me know if there's anything else you'd like me to change in this.

samuelcolvin:
thanks so much.

PR #298, opened 2022-03-09
support python 3.10

### LINKED ISSUE #282: python 3.10 support
greetings @samuelcolvin 
I was trying to test arq on python 3.10.0 and I'm kind of going in circles :snake:  

1. arq uses aioredis 1.3.1 
2. there is a bug on python 3.10.0  (https://bugs.python.org/issue45416) which prevent that version of aioredis to function correctly `        self._cond = asyncio.Condition(lock=Lock())`

what do you think would be the best way to proceed ? 

It seems like trying to push https://github.com/samuelcolvin/arq/pull/259 is the right move however, I haven't check precisely if aioredis 2.x was hit but the python bug (for which there is a workaround) 

I could also ask if aioredis intends to fix the 1.3.1 version, which would sort the 1st blocker I'm facing on this.

let me know your thoughts, thanks !

samuelcolvin:
fix #282

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu]
-        python-version: ['3.7', '3.8', '3.9']
+        python-version: ['3.7', '3.8', '3.9', '3.10']
 
     env:
       PYTHON: ${{ matrix.python-version }}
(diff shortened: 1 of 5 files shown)


JonasKs:
Not sure if you're struggling with this:

```py
@pytest.fixture(scope='session', autouse=True)
def event_loop():
    """
    Reference: https://github.com/pytest-dev/pytest-asyncio/issues/38#issuecomment-264418154
    https://github.com/Kludex/basesqlmodel/blob/d3ee96bf24095f60dd06b8e57b19a86c7b8717c9/tests/conftest.py#L26
    """
    loop = asyncio.get_event_loop_policy().new_event_loop()
    yield loop
    loop.close()
```

PR #301, opened 2022-04-02
Allow unix_socket_path in RedisSettings

lazToum:


tests/requirements-testing.txt
```diff
@@ -1,5 +1,6 @@
 coverage==6.3
 dirty-equals==0.1
+redislite==6.2.805324
```
samuelcolvin, line 3: I don't think we need this, if we do, it's a different discussion and PR.

samuelcolvin (commented):
otherwise looking good.

setup.py
```diff
@@ -55,8 +55,8 @@
         arq=arq.cli:cli
     """,
     install_requires=[
-        'redis>=4.2.0rc2',
-        'click>=6.7',
+        'redis>=4.2.0',
+        'click>=6.7,<8.1.0',
```
samuelcolvin, line 59: is this required?

tests/requirements-testing.txt
lazToum: This is only for testing. I haven't found any easy way to include on the CI workflow a redis-server that is listening on a unix socket.

setup.py
lazToum: The `click>=6.7,<8.1.0`, no, not anymore. There was an issue with black (https://github.com/psf/black/issues/2964). But it looks it's ok now. I removed it. The `redis>=4.2.0` is needed if we include redislite on tests.

tests/requirements-testing.txt
samuelcolvin: Ok, that's fine then.

samuelcolvin:
looks like you have conflicts, hence tests aren't running.

PR #303, opened 2022-04-04
Allow user to configure a default job expiry-extra length

isobelhooper:
Currently, `expires_extra_ms` in arq/connections.py sets a default expiry of 1 day from when a job was expected to start to when it would expire if no expiry was set in `enqueue_job()`.

Our project found we needed that to be longer, so we made it configurable as part of both ArqRedis and Worker, as `.default_job_expiry`.

This PR also moves `expires_extra_ms` to arq/constants.py along with other shared constants.

----

`default_job_expiry` might not be the best name for this option, as it's one of a few different job expiry times, so I'm happy to rename the option as necessary.

### DIFF
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -12,7 +12,7 @@ from redis.asyncio import ConnectionPool, Redis
 from redis.asyncio.sentinel import Sentinel
 from redis.exceptions import RedisError, WatchError
 
-from .constants import default_queue_name, job_key_prefix, result_key_prefix
+from .constants import default_queue_name, expires_extra_ms, job_key_prefix, result_key_prefix
 from .jobs import Deserializer, Job, JobDef, JobResult, Serializer, deserialize_job, serialize_job
 from .utils import timestamp_ms, to_ms, to_unix_ms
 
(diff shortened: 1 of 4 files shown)


arq/connections.py
```diff
@@ -90,13 +88,15 @@ def __init__(
         job_serializer: Optional[Serializer] = None,
         job_deserializer: Optional[Deserializer] = None,
         default_queue_name: str = default_queue_name,
+        default_job_expiry: int = expires_extra_ms,
```
samuelcolvin, line 91:
```suggestion
        expires_extra_ms: int = expires_extra_ms,
```
Might not be the best name, but this will be clearer as a change.

arq/worker.py
```diff
@@ -189,6 +192,7 @@ def __init__(
         max_burst_jobs: int = -1,
         job_serializer: Optional[Serializer] = None,
         job_deserializer: Optional[Deserializer] = None,
+        default_job_expiry: int = expires_extra_ms,
```
samuelcolvin, line 195:
```suggestion
        expires_extra_ms: int = expires_extra_ms,
```
as above.

tests/requirements-linting.txt
```diff
@@ -1,4 +1,4 @@
-black==21.12b0
+black==22.3.0
```
samuelcolvin, line 1: remove this change, it's fixed on main.

samuelcolvin (commented):
otherwise, LGTM.

isobelhooper:
I've changed `default_job_expiry` to `expires_extra_ms` and rebased the changes onto main (which removed the change to the version of `black` too).

samuelcolvin:
great, thanks so much.

PR #311, opened 2022-05-19
Fix recursion while waiting for redis connection

nierob:
`Arq` has ability to retry connections to redis in case of problems.
That is a nice feature, but it crashes after hundreds of attempts
reaching maximal recursion depth. This change modifies re-try algorithm
from recursion to iteration avoiding the limit.

In practice it would be nice to have ability to wait forever as issue
with redis instance should not kill worker process, but that is a
separate change that can be built on top.

Fixes exception when retrying redis connection many times:
  RecursionError: maximum recursion depth exceeded while getting the str of an object


hermansc (approved):
Looks great!

nierob:
Hi, all the failures in CI feels unrelated, should I do something more to get this PR merged?

nierob:
> 3.9 failure looks intermittent, might be worth rerunning.

I have no means of re-running other then changing something and re-pushing. I guess it is related to some permissions.

> If you can fix the problem with click that would be great, otherwise I will.

https://github.com/samuelcolvin/arq/pull/312

samuelcolvin:
Thanks so much.

By the way, you can always re-run CI using `git commit -am bump --allow-empty`.

PR #322, opened 2022-08-12
Added return statement in example `sync_job.py`

pandianmn:
Added return statement in example `sync_job.py`


samuelcolvin:
thanks so much.

PR #323, opened 2022-08-15
Fix/redis ssl support

### LINKED ISSUE #319: Dropped SSL Support for redis connection
I think that the changes introduced here
https://github.com/samuelcolvin/arq/pull/259
somehow broke the support for a complex ssl configuartion, now the `__init__` method from redis requires a boolean as opposed to a boolean or an  `SSLContext` object like it used to. For example that setup does not work anymore:
```python
    ctx = ssl.SSLContext()
    ctx.check_hostname = False
    ctx.verify_mode = ssl.CERT_NONE
    return RedisSettings(
        host=settings.REDIS_HOST,
        port=settings.REDIS_PORT,
        password=settings.REDIS_PASSWORD,
        ssl=ctx,
        conn_timeout=10,
    )
```

Here is an extract of what the `Redis` client init method expects:
```python
    def __init__(
        self,
        *,
        ....
        ssl: bool = False,
        ssl_keyfile: Optional[str] = None,
        ssl_certfile: Optional[str] = None,
        ssl_cert_reqs: str = "required",
        ssl_ca_certs: Optional[str] = None,
        ssl_ca_data: Optional[str] = None,
        ssl_check_hostname: bool = False,
        max_connections: Optional[int] = None,
      ...
    ):
```

nicolas-chaulet:
This is a proposal to fix #319.

When swicthing to redis-py the definition of ssl configuration is not valid anymore and this PR fixes that by matching the way redis-py handles SSL configurations.


samuelcolvin:
great, thanks.

nicolas-chaulet:
You're welcome! Thanks for your great work, happy I could contribute!

PR #326, opened 2022-08-23
add v0.23 history

samuelcolvin:
preparing for release of v0.23 which is just a non-pre copy of v0.23a1.


samuelcolvin:
Version v0.23 is released.

PR #335, opened 2022-08-23
tweak ci a little

samuelcolvin:
hopefully make it a tiny bit faster.


Kludex:
Maybe enabling cache to speed up? https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/

samuelcolvin:
Could do, but probably not worth it given that build is already pretty fast

samuelcolvin:
This is more me messing around with pip-compile and what's possible.

PR #338, opened 2022-08-26
fix: info commands are not allowed inside transactions at least in redis 6.2.3

jerzy-kurowski:


samuelcolvin:
Fine by me, but I don't see a failure with redis 6, see #340.

rolveb:
FYI, The patch worked for using arq against redislabs freetier service (6.2.3). Note that the bug did not affect 6.2.6 version of redis server which e.g. Upstash uses).

samuelcolvin:
Interesting, so it was a very specific version of redis that had this problem.

PR #345, opened 2022-09-18
feat: signal handler to wait for task completion before shutting down

JonasKs:
This is a draft in order to ask for feedback on implementation and to see if this is something that you would consider accepting before I look into tests (which I'm not sure how to do, honestly)

TODO: Write some proper docstrings, write tests and clean up log messages (haven't proof read anything)

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -154,9 +154,10 @@ class Worker:
     :param after_job_end: coroutine function to run after job has ended and results have been recorded
     :param handle_signals: default true, register signal handlers,
       set to false when running inside other async framework
-    :param wait_for_job_completion_on_signal_seconds: time to wait before cancelling tasks after a signal.
+    :param job_completion_wait: time to wait before cancelling tasks after a signal.
       Useful together with `terminationGracePeriodSeconds` in kubernetes,
-      when you want to make the pod complete jobs before shutting down
+      when you want to make the pod complete jobs before shutting down.
+      The worker will not pick new tasks while waiting for shut down.
     :param max_jobs: maximum number of jobs to run at a time
     :param job_timeout: default job timeout (max run time)
     :param keep_result: default duration to keep job results for
@@ -196,7 +197,7 @@ class Worker:
         on_job_end: Optional['StartupShutdown'] = None,
         after_job_end: Optional['StartupShutdown'] = None,
         handle_signals: bool = True,
-        wait_for_job_completion_on_signal_seconds: int = 0,
+        job_completion_wait: int = 0,
         max_jobs: int = 10,
         job_timeout: 'SecondsTimedelta' = 300,
         keep_result: 'SecondsTimedelta' = 3600,
@@ -268,10 +269,14 @@ class Worker:
         self._last_health_check: float = 0
         self._last_health_check_log: Optional[str] = None
         self._handle_signals = handle_signals
-        self._wait_for_job_completion_on_signal_seconds = wait_for_job_completion_on_signal_seconds
+        self._job_completion_wait = job_completion_wait
         if self._handle_signals:
-            self._add_signal_handler(signal.SIGINT, self.handle_sig)
-            self._add_signal_handler(signal.SIGTERM, self.handle_sig)
+            if self._job_completion_wait:
+                self._add_signal_handler(signal.SIGINT, self.handle_sig_wait_for_completion)
+                self._add_signal_handler(signal.SIGTERM, self.handle_sig_wait_for_completion)
+            else:
+                self._add_signal_handler(signal.SIGINT, self.handle_sig)
+                self._add_signal_handler(signal.SIGTERM, self.handle_sig)
         self.on_stop: Optional[Callable[[Signals], None]] = None
         # whether or not to retry jobs on Retry and CancelledError
         self.retry_jobs = retry_jobs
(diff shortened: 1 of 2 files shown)


JonasKs:
TIL you can raise signals from Python - I think this won't be hard to test then. Let me know if interested and I'll add tests 😊

samuelcolvin:
There seem to be two changes here, neither directly related to k8s afaik.

1. Waiting for jobs to finish before stopping. 
2. Not installing the handler on sigint

Is there a particular reason the changes need to come together?

JonasKs:
> Is there a particular reason the changes need to come together?

The normal signal handler would cancel tasks in SIGINT, but I want them to not cancel tasks and rather wait until all tasks are completed. This requires the worker to stop picking tasks, otherwise this could take potentially forever. For this reason, they need to come together. 

I should've written a better README for this PR with my use-case:

In K8s we have a setting called `terminationGracePeriodSeconds`. This is normally 30 seconds, and would be enough for:
* Worker to stop picking jobs
* Worker to finish all current tasks (my job_timeout is 20 seconds, but for those with more, they would have to adjust `terminationGracePeriodSeconds` for this to make sense)
* Worker to gracefully shut 

With the current signal handlers, all tasks are cancelled and the worker instantly shuts down. This is fine (idempotent tasks), but creates some noise. 

So this is honestly just an alternative signal handler for SIGINT, which I would personally only use in k8s - but I agree, they are not directly related to k8s, it's just the use-case origin for me.

arq/worker.py
```diff
@@ -186,6 +186,7 @@ def __init__(
         on_job_start: Optional['StartupShutdown'] = None,
         on_job_end: Optional['StartupShutdown'] = None,
         handle_signals: bool = True,
+        use_k8s_signal_handler: bool = False,
```
JonasKs, line 189: This could be renamed to `cancel_job_on_sigint` or something like that? And then we could generalize the rest.

samuelcolvin: yes agreed.

JonasKs: Thanks! I'll finalize when I'm back from my vacation.

samuelcolvin:
Happy to accept this with tests and docstrings. I think we should avoid `k8s` in variable names etc.

If you want to add a note about k8s and `terminationGracePeriodSeconds` in the docs/docstrings, that's fine by me.

JonasKs:
I will do my best and request review when ready. Probably will be in ~2 weeks. Thanks for the feedback!

samuelcolvin:
great thanks, I think others are keen for a release, so I'll probably make a release next week or as soon as #355 is released, happy to either include this or make another release with this.

JonasKs:
Hey @samuelcolvin , I saw that you wanted to create a new version tomorrow so I sat down and tried to make this more generic.   
I think the solution I have right now is better, and I'll add tests tomorrow (going to bed now). Feel free to review the implementation 😊

arq/worker.py
```diff
@@ -757,8 +764,30 @@ def _add_signal_handler(self, signum: Signals, handler: Callable[[Signals], None
...
         return self.jobs_complete + self.jobs_retried + self.jobs_failed + len(self.tasks)
 
+    async def _wait_for_tasks_to_complete(self) -> None:
+        while self.tasks:
+            await asyncio.sleep(0.1)
+
     def handle_sig(self, signum: Signals) -> None:
         sig = Signals(signum)
+        if self._wait_for_job_completion_on_signal_seconds:
+            logger.info('Setting allow_pick_jobs to `False`')
```
JonasKs, line 774: Redundant probably. 🤔

arq/worker.py
```diff
@@ -757,8 +764,30 @@ def _add_signal_handler(self, signum: Signals, handler: Callable[[Signals], None
...
+            self.allow_pick_jobs = False
+            logger.info(
+                'shutdown on %s ◆ %d jobs complete ◆ %d failed ◆ %d retries ◆ %d ongoing waiting for completion',
```
JonasKs, line 777:
```suggestion
                'shutdown on %s ◆ %d jobs complete ◆ %d failed ◆ %d retries ◆ %d to be completed',
```
```suggestion
                'shutdown on %s ◆ %d jobs complete ◆ %d failed ◆ %d retries ◆ %d ongoing',
```
```suggestion
                'shutdown on %s ◆ %d jobs complete ◆ %d failed ◆ %d retries ◆ %d waiting for completion',
```

I am too tired to find a good sentence 😁 help

arq/worker.py
```diff
@@ -192,6 +196,7 @@ def __init__(
         on_job_end: Optional['StartupShutdown'] = None,
         after_job_end: Optional['StartupShutdown'] = None,
         handle_signals: bool = True,
+        wait_for_job_completion_on_signal_seconds: int = 0,
```
JonasKs, line 199: Feel free to suggest a better name here too.

arq/worker.py
```diff
@@ -153,6 +154,9 @@ class Worker:
     :param after_job_end: coroutine function to run after job has ended and results have been recorded
     :param handle_signals: default true, register signal handlers,
       set to false when running inside other async framework
+    :param wait_for_job_completion_on_signal_seconds: time to wait before cancelling tasks after a signal.
+      Useful together with `terminationGracePeriodSeconds` in kubernetes,
+      when you want to make the pod complete jobs before shutting down
```
JonasKs, line 159: Should probably mention here that it also won't pick new tasks

arq/worker.py
```diff
@@ -757,8 +764,30 @@ def _add_signal_handler(self, signum: Signals, handler: Callable[[Signals], None
...
+                self.jobs_retried,
+                len(self.tasks),
+            )
+            with contextlib.suppress(TimeoutError):
+                self.loop.create_task(
+                    asyncio.wait_for(
+                        self._wait_for_tasks_to_complete(),
+                        self._wait_for_job_completion_on_signal_seconds,
+                    )
+                )
```
JonasKs, line 790: Converting to draft again, I see that this don't actually work as intended. This creates a task, but don't wait for it to complete.

```python
            with contextlib.suppress(TimeoutError):
                logger.info('Waiting %d seconds for tasks to complete', self._wait_for_job_completion_on_signal_seconds)
                wait_for_task = self.loop.create_task(
                    asyncio.wait_for(
                        self._wait_for_tasks_to_complete(),
                        self._wait_for_job_completion_on_signal_seconds,
                    )
                )
                self.loop.run_until_complete(wait_for_task)
```
also don't work. 

I'll have to rewrite to be more like the previous commit I guess, but I don't think I'll make it for this release.

arq/worker.py
samuelcolvin: maybe, but this is good enough for now.

arq/worker.py
samuelcolvin: probably too long, but not enough for me to bother changing it now :smile:.

arq/worker.py
samuelcolvin: :shrug:, I think fine.

arq/worker.py
samuelcolvin: `to be completed` LGTM, not committing to avoid messing with your work.

arq/worker.py
samuelcolvin: Understood, thanks for letting me know.

arq/worker.py
samuelcolvin: `job_completion_wait`??? Should be descriptive enough with a good docstring?

arq/worker.py
JonasKs: Actually, I think I've made it. Give me one sec. I'll fix other feedback too.

samuelcolvin:
okay, let me know when it's ready to review, no special hurry, I'm in the office most of the day.

JonasKs:
I've rebased and finished up the PR, @samuelcolvin . Thanks for your patience and helping me out with the namings 😊 

Pipelines seems to break, but I'm not sure why, and they don't seem related to my changes. Maybe some timing issues again? Seems to be the same test every time though 🤔

arq/worker.py
```diff
@@ -153,6 +154,10 @@ class Worker:
     :param after_job_end: coroutine function to run after job has ended and results have been recorded
     :param handle_signals: default true, register signal handlers,
       set to false when running inside other async framework
+    :param job_completion_wait: time to wait before cancelling tasks after a signal.
+      Useful together with `terminationGracePeriodSeconds` in kubernetes,
```
samuelcolvin, line 158:
```suggestion
      Useful together with ``terminationGracePeriodSeconds`` in kubernetes,
```
until we move to markdown, this is still rst.

arq/worker.py
```diff
@@ -192,6 +197,7 @@ def __init__(
         on_job_end: Optional['StartupShutdown'] = None,
         after_job_end: Optional['StartupShutdown'] = None,
         handle_signals: bool = True,
+        job_completion_wait: int = 0,
```
JonasKs, line 200: I see you've used `'SecondsTimedelta'` other places, but this was not accepted by mypy and `asyncio.wait_for()`.

samuelcolvin: I think this is clear.

samuelcolvin:
:disappointed: looks like we need to look at those intermittent errors, but let's do that once we get this released.

JonasKs:
Thank you so much! 😊

samuelcolvin:
> Thank you so much! :blush: 

No, thank you for your contribution.

I know arq has become a bit "unloved" as I've been concentrating on pydantic, so really helpful to get contributions like this from others. I've get back to it when pydantic V2 is released.

> I also cannot reproduce locally (macos, 3.11). I've run the tests 100 times now, no failures.

I think it must be a timing thing since I guess CI runs on shared hardware with lots of other jobs, some tasks can take significantly longer than they ever would locally, especially since redis is running in docker, maybe even docker in on a different host.

JonasKs:
> I know arq has become a bit "unloved" as I've been concentrating on pydantic, so really helpful to get contributions like this from others. I'll get back to it when pydantic V2 is released.

You might not commit code to it as much as you once did, but I think we all agree that it's "loved" as long as you are reviewing contributions once every few months 😊 It's totally fair to rely on the community, especially when it's very stable and the code is easy to contribute to.

Issue #349, opened 2022-09-27
Do not log job result


iamlikeme:
When a job successfully completes arq logs the job result:

https://github.com/samuelcolvin/arq/blob/1495be6c234509df1dfdc0db620388d909c8cdb8/arq/worker.py#L588

This is not always desirable. Results may contain sensitive information that should not end up in logs. I believe this may be a problem for many other users of arq.

Currently there is no good way to disable this behavior. I considered changing log level for arq's logger to WARNING, but then I would miss other useful logged with the INFO level, e.g. job retires, cancellations, etc. I see two possible solutions:

1. Remove `result_str` from the log
2. Make the behavior configurable, e.g. by via WorkerSettings

I would be glad to contribute a PR if one of the above proposals or some other solution is approved.

JonasKs:
I would very much like these logs and vote for a solution to configure it. Health-check-alike logs like these that suddenly disappears might also break dashboards etc.

iamlikeme:
@JonasKs Agree, I also prefer to make it configurable. I submitted implementation in #352.

samuelcolvin:
👍 thank you, I agree it's good to make it configurable.

I'll review next week.

PR #350, opened 2022-09-27
Do not log job result

iamlikeme:
It is not always desirable to include job result in logs.

This PR is one of the possible solutions to #349, where job result is simply removed from the logs.
An alternative would be to make the behavior configurable.


samuelcolvin:
this behaviour should definitely be configurable, and default to include the result.

samuelcolvin:
used #352 instead.

Issue #351, opened 2022-10-09
Timezone support for cron jobs


iamlikeme:
A job scheduled to run at 12 o'clock will be started when it is 12 o'clock *local time* on the host where the worker is running (because of how `datetime.now()` works). This means that when a job is started depends on where the host is located and how it is configured. This does not seem optimal.

Timezone support for cron jobs was already requested in #195 and #251, both rejected. The suggested workaround was to schedule a job frequently and then check during job execution if it is the right time. The reasoning was (ref https://github.com/samuelcolvin/arq/issues/251#issuecomment-852883654):

> my instinct is that this isn't possible since we using unix timestamps to denote when jobs should start.

I am trying to understand why this is a problem. I assume that "using unix timestamps to denote when jobs should start" refers to the job "score" (as stored in Redis) being a unix timestamp. The score is calculated from an optional datetime argument `_defer_until` passed to `enqueue_job`:

https://github.com/samuelcolvin/arq/blob/1495be6c234509df1dfdc0db620388d909c8cdb8/arq/connections.py#L153-L154

From what I can see passing time zone "aware" datetime objects `to_unix_ms` works just fine:

```python
from datetime import datetime
from zoneinfo import ZoneInfo
from arq.utils import to_unix_ms

# My local time as naive datetime
to_unix_ms(datetime(2022, 10, 9, 20))
# 1665338400000

# I happen to live in Norway, so adding my local time zone does not change the result
to_unix_ms(datetime(2022, 10, 9, 20, tzinfo=ZoneInfo("Europe/Oslo")))
# 1665338400000

# My local time convered to a different timezone also produces the same unix timestamp
to_unix_ms(datetime(2022, 10, 9, 14, tzinfo=ZoneInfo("America/New_York")))
#1665338400000
```

So it seems to me that adding time zone support for cron jobs would mean making sure that `arq.cron.next_cron()` returns an "aware" datetime object with a time zone specified by user (e.g. as a Worker attribute). Do you think it would work? If yes, I believe it is something I could contribute.

iamlikeme:
I submitted an implementation of the timezone support. As you can see it requires very few changes.

PR #354, opened 2022-10-16
Add timezone support for cron jobs

iamlikeme:
This changeset ensures that calculations of execution time for cron jobs are done on timezone-aware datetime objects. Timezone can be specified by a new `Worker` attribute `timezone` which defaults to the system timezone (which preserves the current behavior).

Fixes: #351


samuelcolvin:
This is great, thanks so much.

samuelcolvin:
I spent a long time trying to get it to yield an incorrect timezone e.g. using pytz, but it seems your logic is correct. 🎉

PR #355, opened 2022-11-02
Adds after_job_end hook

AngellusMortis:
Adds an additional hook for after the job completes. Useful for doing something _after_ the results for the job has been recorded.

My specific use case is I am adding New Relic integration into arq for our projects. Being able to record the success/failure, queue time, etc. of a job directly to New Relic is very useful.

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -150,6 +150,7 @@ class Worker:
     :param on_shutdown: coroutine function to run at shutdown
     :param on_job_start: coroutine function to run on job start
     :param on_job_end: coroutine function to run on job end
+    :param after_job_end: coroutine function to run after job has ended and results have been recorded
     :param handle_signals: default true, register signal handlers,
       set to false when running inside other async framework
     :param max_jobs: maximum number of jobs to run at a time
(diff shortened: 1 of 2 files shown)


JonasKs:
What happens if this task has an exception? `on_job_end` failure acts like the task itself failed, and I don’t believe that’s what you’d like in this scenario.

AngellusMortis:
`on_job_end` and `on_job_start` do not handle exception either. I think all 3 of them should, but I think that is out of scope of this PR.

AngellusMortis:
Probably more specifically, it would be nice to see something like this:

```python
async def run_job(...):
    ....
    
    async def job_wrapper():
        await on_job_start(ctx)
        await function.coroutine(ctx, *args, **kwargs)
        await on_job_end(ctx)
     
    await asyncio.shield(before_job_start(ctx))
    await job_wrapper()
    await await asyncio.shield(self.finish_job())
    await asyncio.shield(after_job_end(ctx))
```

This would allow for:
* optional before/after hooks for reporting/etc. that can fail without affecting the outcome of the task
* a required before/after job hooks that are _required_ to succeed for the job to succeed.

samuelcolvin:
@AngellusMortis LGTM, could you resolve conflicts.

AngellusMortis:
Fixed.

samuelcolvin:
thanks so much.

PR #362, opened 2022-11-19
Fix race condition causing incorrect status not_found

### LINKED ISSUE #329: Status statistically return not found for job that just ended
When using ARQ as our job management system - we encountered a case where job status sampling returned "not found" even though the job existed. After investigating the issue, we found a race between the job's [status](https://github.com/samuelcolvin/arq/blob/master/arq/jobs.py#L135) function and the worker's [finish_job](https://github.com/samuelcolvin/arq/blob/master/arq/worker.py#L609) function.
The finish_job function updates the job's status on Redis from in progress to done using a single transaction. The race is created by the status function, which first checks whether the job is done on the Redis and, if it's not the case, then checks whether the job is in progress on the Redis. The problem occurs when the finish_job transaction is executed between the first status sampling and the second one.
This race can only happen once per job, so our current patch solution is to recheck the status after receiving a not found answer, and if the answer is still not found, then it's not the described race. We would appreciate a more stable solution inside the library. Thank you!

iamlikeme:
Checking job status consited of three separate calls to Redis, where the first one checked if result is present and the second checked if job is in progress. If a job got completed between these two calls Job.status() would erroneously return status not_found.

The solution is to make all the calls to Redis in one transaction.

Fixes #329

### DIFF
--- a/arq/jobs.py
+++ b/arq/jobs.py
@@ -137,15 +137,20 @@ class Job:
         """
         Status of the job.
         """
-        if await self._redis.exists(result_key_prefix + self.job_id):
+        async with self._redis.pipeline(transaction=True) as tr:
+            tr.exists(result_key_prefix + self.job_id)  # type: ignore[unused-coroutine]
+            tr.exists(in_progress_key_prefix + self.job_id)  # type: ignore[unused-coroutine]
+            tr.zscore(self._queue_name, self.job_id)  # type: ignore[unused-coroutine]
+            is_complete, is_in_progress, score = await tr.execute()
+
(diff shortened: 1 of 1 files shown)


iamlikeme:
I had a hard time writing good tests for this change. I could reproduce the issue by patching some implementation details in `Worker.finish_job()` and `Job.status()` but after fixing the problem the patches no longer made sense (because implementation changed). So there is no proof that issue #329 is fixed, but I believe it is reasonable to believe it is.

samuelcolvin:
thanks so much.

PR #364, opened 2022-11-20
Raise ResultNotFound when Job.result() finds no job and no result

iamlikeme:
Job.result() assumed that a result will eventually show up in Redis. This assumption does not hold for jobs which are not in the queue and jobs which do not keep result. In such cases Job.result() would hang forever (or until the optional timeout is reached).

This PR changes Job.result() so that it raises a new exception ResultNotFound, rather than waiting forever, if result is not found and the job is not in the queue.

### DIFF
--- a/arq/jobs.py
+++ b/arq/jobs.py
@@ -18,6 +18,10 @@ Serializer = Callable[[Dict[str, Any]], bytes]
 Deserializer = Callable[[bytes], Dict[str, Any]]
 
 
+class ResultNotFound(Exception):
+    pass
+
+
 class JobStatus(str, Enum):
     """
     Enum of job statuses.
@@ -96,15 +100,21 @@ class Job:
             poll_delay = pole_delay
 
         async for delay in poll(poll_delay):
-            info = await self.result_info()
-            if info:
-                result = info.result
+            async with self._redis.pipeline(transaction=True):
+                v = await self._redis.get(result_key_prefix + self.job_id)
+                s = await self._redis.zscore(self._queue_name, self.job_id)
+
+            if v:
+                info = deserialize_result(v, deserializer=self._deserializer)
                 if info.success:
-                    return result
-                elif isinstance(result, (Exception, asyncio.CancelledError)):
-                    raise result
+                    return info.result
+                elif isinstance(info.result, (Exception, asyncio.CancelledError)):
+                    raise info.result
                 else:
-                    raise SerializationError(result)
+                    raise SerializationError(info.result)
+            elif s is None:
+                raise ResultNotFound()
+
             if timeout is not None and delay > timeout:
                 raise asyncio.TimeoutError()
 
--- a/tests/test_jobs.py
+++ b/tests/test_jobs.py
@@ -7,7 +7,15 @@ from dirty_equals import IsNow
 from arq import Worker, func
 from arq.connections import ArqRedis, RedisSettings, create_pool
 from arq.constants import default_queue_name, in_progress_key_prefix, job_key_prefix, result_key_prefix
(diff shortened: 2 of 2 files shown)


arq/jobs.py
```diff
@@ -18,6 +18,10 @@
 Deserializer = Callable[[bytes], Dict[str, Any]]
 
 
+class ResultNotFound(Exception):
```
samuelcolvin, line 21:
```suggestion
class ResultNotFound(RuntimeError):
```

arq/jobs.py
```diff
@@ -96,15 +100,21 @@ async def result(
...
-                elif isinstance(result, (Exception, asyncio.CancelledError)):
-                    raise result
+                    return info.result
+                elif isinstance(info.result, (Exception, asyncio.CancelledError)):
+                    raise info.result
                 else:
-                    raise SerializationError(result)
+                    raise SerializationError(info.result)
+            elif s is None:
+                raise ResultNotFound()
```
samuelcolvin, line 116: can you include a clear message here.

samuelcolvin (commented):
Looks like a good idea to me, just a few things to change to change.

Also can you add some info about this to `result` docstring.

arq/jobs.py
iamlikeme: Done

iamlikeme:
I updated docstring and error message. If you want to change wording feel free to push directly to this branch.

iamlikeme:
The new `ResultNotFound` exception may now occur in `Job.abort()`. When it does, we don't know whether the job was cancelled or not (e.g. the job might have been cancelled, but if it does not keep result we wouldn't know). There are several options for what we can do with it:

* not handle it, i.e. let `Job.abort()` raise `ResultNotFound`
* catch and return `False` (i.e. job wasn't cancelled)
* catch and return `True` (i.e. job was cancelled)
* catch and return something else, e.g. `None` - this would be a breaking change

The first two options seem most sensible to me and I chose to catch and return `False`. Let me know if you'd rather have it changed.

arq/jobs.py
```diff
@@ -82,8 +86,10 @@ async def result(
         self, timeout: Optional[float] = None, *, poll_delay: float = 0.5, pole_delay: float = None
     ) -> Any:
         """
-        Get the result of the job, including waiting if it's not yet available. If the job raised an exception,
-        it will be raised here.
+        Get the result of the job or, if the job raised an exception, reraise it.
+
+        This function waits for the result if it's not yet available and the job is
+        present in the queue. Otherwise ResultNotFound is raised.
```
samuelcolvin, line 92:
```suggestion
        present in the queue. Otherwise ``ResultNotFound`` is raised.
```

samuelcolvin:
Great, thanks so much.

> The first two options seem most sensible to me and I chose to catch and return `False`

Agreed, I think returning `False` is most correct since it's matches the docstring: `True if the job aborted properly, False otherwise`

shahakL:
Hi @iamlikeme! Quick question about this PR, I noticed this part in `Job.result()`:
```python
async with self._redis.pipeline(transaction=True):
    v = await self._redis.get(result_key_prefix + self.job_id)
    s = await self._redis.zscore(self._queue_name, self.job_id)
```

A transaction is created but not used for the two queries; is that on purpose? Couldn't it create a possible false ResultNotFound if the job is done and removed from the queue after the first line but before the second?

iamlikeme:
Hi @shahakL! You are completely right. What a great catch, thank you! The fix is in #374.

shahakL:
@iamlikeme Thank you for your contributions!

PR #366, opened 2022-11-21
connections: fix pipeline usage for exists command

utkarshgupta137:
Potential fix for: https://github.com/redis/redis-py/issues/2407

### DIFF
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -145,7 +145,7 @@ class ArqRedis(BaseRedis):
 
         async with self.pipeline(transaction=True) as pipe:
             await pipe.watch(job_key)
-            if any(await asyncio.gather(pipe.exists(job_key), pipe.exists(result_key_prefix + job_id))):
+            if await pipe.exists(job_key, result_key_prefix + job_id):
                 await pipe.reset()
                 return None
 


chayim:
@utkarshgupta137 Don't know if you saw the python 3.7 (redis 5) CI fail here. Tagged you because...

utkarshgupta137:
> @utkarshgupta137 Don't know if you saw the python 3.7 (redis 5) CI fail here. Tagged you because...

I saw the failure, but I'm unable to figure out the issue from a brief look. I don't have the time to look into it, so I figured that the library's maintainers would be able to check why it is failing.

samuelcolvin:
I think there are some intermittent failures due to timing issues, I've rerun the job in the hope it passes.

samuelcolvin:
Please can someone explain why this can't be run as part of the pipeline? It would surely be cleaner and faster to include this in the pipeline.

utkarshgupta137:
> Please can someone explain why this can't be run as part of the pipeline? It would surely be cleaner and faster to include this in the pipeline.

You can run it in pipeline. The problem is that you're calling the same pipeline twice in parallel (asyncio.gather). You should either call the pipeline once or not use pipeline if calling in parallel.

utkarshgupta137:
> makes sense, then I would suggest we drop the `asyncio.gather` do the queries sequentially but using the pipeline.

Even better: Since the exists command can accept multiple keys, we can check both of them in a single command. Not sure if you still need a pipeline with that, but I've left it since you're using a transaction.

JonasKs:
Pipeline does not test against release candidate on [redis-py](https://github.com/redis/redis-py), have you tried locally and ensure this works? If not I can check later / we should add a pipeline.

samuelcolvin:
Thanks so much.

Is anything else required to support `redis==4.4.0`?

JonasKs:
Installed arq from master in one of my projects and all tests + functionality seems to run fine here now. Thanks a lot all of you 😊

PR #368, opened 2022-12-04
Use instance's default queue for queued_jobs default

phy1729:
I think it would be less surprising for `queued_jobs` to default to the instance's default rather than the global default.

### DIFF
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -195,10 +195,12 @@ class ArqRedis(BaseRedis):
         jd.score = score
         return jd
 
-    async def queued_jobs(self, *, queue_name: str = default_queue_name) -> List[JobDef]:
+    async def queued_jobs(self, *, queue_name: Optional[str] = None) -> List[JobDef]:
         """
         Get information about queued, mostly useful when testing.
         """
+        if queue_name is None:
+            queue_name = self.default_queue_name
         jobs = await self.zrange(queue_name, withscores=True, start=0, end=-1)
         return await asyncio.gather(*[self._get_job_def(job_id, int(score)) for job_id, score in jobs])
 


arq/connections.py
```diff
@@ -195,10 +195,12 @@ async def _get_job_def(self, job_id: bytes, score: int) -> JobDef:
         jd.score = score
         return jd
 
-    async def queued_jobs(self, *, queue_name: str = default_queue_name) -> List[JobDef]:
+    async def queued_jobs(self, *, queue_name: Optional[str] = None) -> List[JobDef]:
```
JonasKs, line 198: You can change the behavior (like you have) without changing this type annotation. 

The change here is wrong.

JonasKs (changes requested):

arq/connections.py
phy1729: While mypy doesn't warn with `queue_name: str`, `reveal_type(queue_name)` shows `arq/connections.py:202: note: Revealed type is "Union[builtins.str, None]"` and mypy does warn in strict mode. From a quick grep using `: Optional[...] = None` is more common than `: ... = None` so it would not even be matching style.

JonasKs: Right now your implementation is based off `if queue_name is None`, but your type annotation says it cannot be `None`.  
If you look at `enqueue_job`'s implementation you see what I mean.

JonasKs: Ah shit, I read this the second I woke up and apparently are color blind. My bad.. Your implementation is correct.

JonasKs (approved):
Looks good to me. Sorry about my tired eyes causing the confusion 😁

samuelcolvin:

PR #370, opened 2022-12-06
Prevent worker getting stuck in terminating state

### LINKED ISSUE #369: Workers stuck in terminating state when 'job_completion_wait' is specified
Currently the workers will be stuck in a termination state when the handle_sig_wait_for_completion is activated, causing ineffective k8s pod lifecycles.

This appears to be happening since the '_sleep_until_tasks_complete' function is checking len() of self.tasks, but the deletion of done task ids is nested under the if self.allow_pick_jobs. reference: https://github.com/samuelcolvin/arq/blob/main/arq/worker.py#L388

Have tried a couple things:
1. tabbing over the task deletion out from the if self.allow_pick_jobs check
2. having '_sleep_until_tasks_complete' read from self.job_tasks, which are being deleted during finally block in 'run_job'. reference: https://github.com/samuelcolvin/arq/blob/main/arq/worker.py#L583

For both attempted fixes, have been met with asyncio.exceptions.CancelledError errors.

A fix that seems to be working, iterate tasks during '_sleep_until_tasks_complete' while loop and search for not done.

Proposed fix:

Existing code:

    async def _sleep_until_tasks_complete(self) -> None:  
        """
        Sleeps until all tasks are done. Used together with asyncio.wait_for()
        """
        while len(self.tasks):
            await asyncio.sleep(0.1)

Recommended fix:

    async def _sleep_until_tasks_complete(self) -> None:  
        """
        Sleeps until all tasks are done. Used together with asyncio.wait_for()
        """
        while len([task for task in self.tasks.values() if not task.done()]):
            await asyncio.sleep(0.1)

Is this a change you would support if a PR is made, or is there a better way around this issue?

JonasKs:
First commit proves the bug (see the pipeline failures [here](https://github.com/samuelcolvin/arq/actions/runs/3628698559))  
Second commit contains the fix. 

Closes #369

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -382,14 +382,14 @@ class Worker:
 
             await self.start_jobs(job_ids)
 
-            if self.allow_abort_jobs:
-                await self._cancel_aborted_jobs()
+        if self.allow_abort_jobs:
+            await self._cancel_aborted_jobs()
 
-            for job_id, t in list(self.tasks.items()):
-                if t.done():
-                    del self.tasks[job_id]
-                    # required to make sure errors in run_job get propagated
-                    t.result()
+        for job_id, t in list(self.tasks.items()):
+            if t.done():
+                del self.tasks[job_id]
+                # required to make sure errors in run_job get propagated
+                t.result()
 
         await self.heart_beat()
 
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -116,19 +116,23 @@ async def test_worker_signal_completes_job_before_shutting_down(caplog, arq_redi
     async def sleep_job(ctx, time):
         await asyncio.sleep(time)
 
-    await arq_redis.enqueue_job('sleep_job', 0.2, _job_id='short_sleep')  # should be cancelled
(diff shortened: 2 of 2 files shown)


samuelcolvin (commented):
LGTM, but I don't have time right now to dig deep into this.

@mernmic can you confirm if this fixes your problem?

mernmic:
@JonasKs @samuelcolvin yes this has fixed the issue.  thanks!

mernmic:
> LGTM, but I don't have time right now to dig deep into this.
> 
> @mernmic can you confirm if this fixes your problem?

@JonasKs @samuelcolvin yes this has fixed the issue. thanks!

samuelcolvin:
Anything else, or shall we create a patch release?

mernmic:
> Anything else, or shall we create a patch release?

This addresses any issues we've been seeing. We'll more than happily further test the patch once released.

JonasKs:
Thanks @samuelcolvin, sorry for the mess.

samuelcolvin:

Issue #372, opened 2022-12-08
Unable to get a single queued job by id?


VincentRPS:
From the docs, it seems like you cannot get a single job from it's id alone? Unless I'm missing something, adding a feature for this would be perferable.

PR #377, opened 2022-12-15
docs: add documentation on how to retrieve running jobs

### LINKED ISSUE #376: How to get jobs in order to abort
Hi and thanks for this work!

In order to abort some deferred jobs, I need to get an instance of `arq.jobs.Job` but I don't find how to get jobs.
I'm using FastAPI with an endpoint to enqueue job and another to abort job, so I don't have the job instance returns by `enqueue_job()`.

How can I do that?

Thanks

JonasKs:
Related to #376 (and #372)

### DIFF
--- a/docs/examples/job_ids.py
+++ b/docs/examples/job_ids.py
@@ -2,6 +2,8 @@ import asyncio
 
 from arq import create_pool
 from arq.connections import RedisSettings
+from arq.jobs import Job
+
 
 async def the_task(ctx):
     print('running the task with id', ctx['job_id'])
@@ -35,6 +37,14 @@ async def main():
     print(job4)
     """
     >  None
+    """
+
+    # you can retrieve jobs by using arq.jobs.Job
+    await redis.enqueue_job('the_task', _job_id='my_job')
+    job5 = Job(job_id='my_job', redis=redis)
+    print(job5)
+    """
+    <arq job my_job>
     """
 
 class WorkerSettings:


JonasKs:
Tried to re-run tests many times now, they refuse to pass.. I'll try to look into it ASAP.

JonasKs:
Hey @samuelcolvin, seems like `pytest-asyncio==0.20.3` should fix broken pipelines, according to https://github.com/pytest-dev/pytest-asyncio/issues/460. I tried to uprev these dependencies locally, but never used `pip-tools` and `pip-compile` and were not successful in generating dependencies that does not contain conflicts between `pyproject.txt` and `testing.txt`. Unfortunately I don't have time to look into this more today. 

Another potential good thing to do is to use `get_running_loop()` over `get_event_loop()`, as per [documentation](https://docs.python.org/release/3.10.9/library/asyncio-eventloop.html#asyncio.get_event_loop).

> Because this function has rather complex behavior (especially when custom event loop policies are in use), using the [get_running_loop()](https://docs.python.org/release/3.10.9/library/asyncio-eventloop.html#asyncio.get_running_loop) function is preferred to [get_event_loop()](https://docs.python.org/release/3.10.9/library/asyncio-eventloop.html#asyncio.get_event_loop) in coroutines and callbacks.

image

JonasKs:
Seems like new PRs magically got pipelines running again 🤷‍♂️😁

Kludex (approved):

PR #378, opened 2022-12-15
feat: add job_id to JobDef, closing #376

### LINKED ISSUE #376: How to get jobs in order to abort

JonasKs:
This probably need a good review. I looked at the discussion in #198, which had a good reason to use `.get('id')` over `d['id']`. 

Close #376

### DIFF
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -193,6 +193,7 @@ class ArqRedis(BaseRedis):
         assert v is not None, f'job "{key}" not found'
         jd = deserialize_job(v, deserializer=self.job_deserializer)
         jd.score = score
+        jd.job_id = job_id.decode()
         return jd
 
     async def queued_jobs(self, *, queue_name: Optional[str] = None) -> List[JobDef]:
--- a/arq/jobs.py
+++ b/arq/jobs.py
@@ -47,6 +47,7 @@ class JobDef:
     job_try: int
     enqueue_time: datetime
     score: Optional[int]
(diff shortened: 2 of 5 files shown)


JonasKs:
Hi @gerazenobi 😊
It needs a review from @samuelcolvin, who's a busy man these days (pydantic v2 is around the corner). I have no estimate, but I'd assume some time after v2 is out 😊

gerazenobi:
@JonasKs thanks for the quick answer ! 

Just to confirm: there is no workaround to get from queued job to actual job correct ?  (without doing any custom tracking on my end)

gerazenobi:
@JonasKs indeed, however [I need to be able to enqueue multiple jobs of the "same nature"](https://github.com/samuelcolvin/arq/issues/394) hence I can't use custom_ids unfortunately: once an id is used then we can't delete the job nor re-enqueue the job again.

rossmacarthur:
Would love to see this added 🙏

KShah707:
This would be very helpful!

samuelcolvin:
Thanks so much and sorry for the delay in reviewing and merging.

samuelcolvin:

PR #379, opened 2022-12-15
refactor: refactor all asserts into raise <exception>, close #371

### LINKED ISSUE #371: It would be cool to get rid of assert in production code
I suddenly discovered that arq production code (e.g., `connection_pool`) uses asserts.
This is unsafe because of the -O, -OO options in the python interpreter  https://docs.python.org/3/using/cmdline.html#cmdoption-O, as well as the [PYTHONOPTIMIZE](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONOPTIMIZE) environment variable.
I might suggest using instead:
```python
assert condition, message_body.
```
something like this:
```python
If not condition:
    raise AssertionError(message_body)
```

JonasKs:
Close #371 

Can be closed if #371 is not accepted.

### DIFF
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -50,7 +50,8 @@ class RedisSettings:
     @classmethod
     def from_dsn(cls, dsn: str) -> 'RedisSettings':
         conf = urlparse(dsn)
-        assert conf.scheme in {'redis', 'rediss', 'unix'}, 'invalid DSN scheme'
+        if conf.scheme not in {'redis', 'rediss', 'unix'}:
+            raise RuntimeError('invalid DSN scheme')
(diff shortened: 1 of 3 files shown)


samuelcolvin:
> Why is this PR not accepted?

Because I've been very busy. My mistake.

I think we should use `ValueError` or `TypeError` for many of these as they're the wrong value or the wrong type.

JonasKs:
Cheers Samuel, I'll look over the errors when I find time. Probably next week or so, I'm gone for a work trip. 

If you want to fork my fork and contribute fixes to it, you're welcome to do so @WillemDeGroef.

samuelcolvin:

PR #382, opened 2022-12-17
chore: update dependencies, fixing tests

JonasKs:
**EDIT April 4th**: Seems like this PR is no longer needed, as pipelines have fixed themselves. 

As mentioned in https://github.com/samuelcolvin/arq/pull/377#issuecomment-1354594601: 

> seems like pytest-asyncio==0.20.3 should fix broken pipelines, according to https://github.com/pytest-dev/pytest-asyncio/issues/460. 

I also added 

> Another potential good thing to do is to use get_running_loop() over get_event_loop(), as per [documentation](https://docs.python.org/release/3.10.9/library/asyncio-eventloop.html#asyncio.get_event_loop).

But I can remove this if you'd like. 

I can rebase the other PRs I have open if this gets merged to make pipelines pass. 😊


arq/cli.py
```diff
@@ -60,7 +60,7 @@ async def watch_reload(path: str, worker_settings: 'WorkerSettingsType') -> None
     except ImportError as e:  # pragma: no cover
         raise ImportError('watchfiles not installed, use `pip install watchfiles`') from e
 
-    loop = asyncio.get_event_loop()
+    loop = asyncio.get_running_loop()
```
JonasKs, line 63: Pipelines fixed themselves, so I don't think we need to merge this PR. 

This is somewhat still relevant, but also not really needed. Feel free to close. 

> Another potential good thing to do is to use get_running_loop() over get_event_loop(), as per [documentation](https://docs.python.org/release/3.10.9/library/asyncio-eventloop.html#asyncio.get_event_loop).

Issue #380, opened 2022-12-16
Datetime in next_cron func doesn't have timezone information 


TyrPhoonn:
I have one issue at line 47 in `cron.py` return next_datetime values without timezone information.
I think it will be happen when your frequency_time greater than now and next_cron will move to next year.
You can try with current time and next_cron is any day of next year.

You can use this code to replace while loop at line 43
```
while True:
    next_dt = _get_next_dt(dt, options)
    if next_dt is None:
        return dt.replace(tzinfo=datetime.now().astimezone().tzinfo)
    dt = next_dt
```

JonasKs:
Maybe @iamlikeme who built tz aware cron jobs has any insight here? 😊 I won't be able to look into this for a while.

JonasKs:
PRs with tests always welcome 🙏

iamlikeme:
Thanks @TyrPhoonn for reporting the bug, and thanks @JonasKs for pinging me :+1: I submitted a PR with a bugfix.

PR #383, opened 2022-12-18
Fix: timezone info occasionally removed from cron job execution time

iamlikeme:
This PR fixes a bug where timezone information was occasionally removed from execution time of cron jobs. The bug affected cron jobs which used the month-rule and when month was incremented due to that rule.

Since #354 all cron calculations are done on timezone aware datetimes. However, the tests for `next_cron` function were not updated to include timezone in all test cases, which is how the bug slipped through. The tests are updated now to always use timezone.

Fixes: #380

### DIFF
--- a/arq/cron.py
+++ b/arq/cron.py
@@ -66,9 +66,9 @@ def _get_next_dt(dt_: datetime, options: Options) -> Optional[datetime]:  # noqa
             micro = max(dt_.microsecond - options.microsecond, 0)
             if field == 'month':
                 if dt_.month == 12:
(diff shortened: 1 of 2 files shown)


JonasKs:
Tests has been fixed in #382, so don't worry about the pipelines failing for some Python versions.

JonasKs (approved):

iamlikeme:
Ready for merging :pray:

nierob (approved):

samuelcolvin:
thanks so much, sorry for the long delay.

samuelcolvin:

Issue #386, opened 2023-01-17
Allow workers to specify redis Retry Helpers settings


mernmic:
Currently the retry settings conn_retries, conn_timeout, conn_retry_delay are only applied to the initial connection and any connection issues while running, eg: temporary network hiccup, disconnected by host, will cause the worker to crash.

Proposing extending for the following settings:
    retry: Retry instance with a Backoff strategy and the max number of retries
    retry_on_error: list of Exceptions to retry on
    retry_on_timeout: if True, retry on TimeoutError only

redis-py documentation on retry helpers:
https://redis-py.readthedocs.io/en/stable/retry.html

mernmic:
example stack trace:
``` 
File "/project_dir/venv/lib/python3.9/site-packages/arq/worker.py", line 868, in run_worker
    worker.run()
  File "/project_dir/venv/lib/python3.9/site-packages/arq/worker.py", line 301, in run
    self.loop.run_until_complete(self.main_task)
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/project_dir/venv/lib/python3.9/site-packages/arq/worker.py", line 354, in main
    await self._poll_iteration()
  File "/project_dir/fc_tq/utils/arq_utils.py", line 181, in custom_poll_iteration
    job_ids = await worker.pool.zrangebyscore(
  File "/project_dir/venv/lib/python3.9/site-packages/redis/asyncio/client.py", line 487, in execute_command
    return await conn.retry.call_with_retry(
  File "/project_dir/venv/lib/python3.9/site-packages/redis/asyncio/retry.py", line 62, in call_with_retry
    await fail(error)
  File "/project_dir/venv/lib/python3.9/site-packages/redis/asyncio/client.py", line 476, in _disconnect_raise
    raise error
  File "/project_dir/venv/lib/python3.9/site-packages/redis/asyncio/retry.py", line 59, in call_with_retry
    return await do()
  File "/project_dir/venv/lib/python3.9/site-packages/redis/asyncio/client.py", line 463, in _send_command_parse_response
    return await self.parse_response(conn, command_name, **options)
  File "/project_dir/venv/lib/python3.9/site-packages/redis/asyncio/client.py", line 505, in parse_response
    response = await connection.read_response()
[...]

PR #387, opened 2023-01-17
Extend RedisSettings to include redis Retry Helper settings

mernmic:
proposed fix for https://github.com/samuelcolvin/arq/issues/386


arq/connections.py
```diff
@@ -47,6 +47,10 @@ class RedisSettings:
     sentinel: bool = False
     sentinel_master: str = 'mymaster'
 
+    retry_on_timeout: bool = False
+    retry_on_error: Optional[List[Any]] = None
+    retry: Optional[Any] = None
```
JonasKs, line 52: This isn't an `Any`, but rather a `redis.Retry`, right?

JonasKs (changes requested):

arq/connections.py
mernmic: yup, fixed.

JonasKs:
The tests has been fixed in #382, so don’t worry about those failing pipelines on some Python versions for now.  

As you can see from the codecov bot above, this repo strive to have 100% codecov. Would love if you added tests for these settings.   
If you’re unsure how to proceed, please don’t hesitate to ask. 😊

mernmic:
@JonasKs, added some tests which improved coverage.  Still one partial, but might be from testing failures?

JonasKs:
I'll try to review again this weekend / early next week 🙇‍♂️

arq/connections.py
```diff
@@ -47,6 +48,10 @@ class RedisSettings:
     sentinel: bool = False
     sentinel_master: str = 'mymaster'
 
+    retry_on_timeout: bool = False
+    retry_on_error: Optional[List[Any]] = None
```
JonasKs, line 52: Wouldn't this be `Optional[List[Exception]]`?

JonasKs (approved):
Looks good to me! 😊

arq/connections.py
mernmic: Yup, fixed.  thanks for catching.

mernmic:
> Looks good to me! 😊

Thanks for taking a look., just pushed a change for the Exception type.

We've been using the above changes in our systems and everything has been working as expected so far.

A heads up I'll be offline from the first few weeks of February, so if anything comes up that's the reason for any radio silence on my end.

Thanks again!

JonasKs:
Thanks for the PR and for testing in your environment! 
Enjoy your offline time☺️     

I'd also recommend double checking that maintainers can edit your PR, in case Samuel finds time to review/release while you're gone. 😊

maximgrigorov:
Hi there,

is it possible to merge this fix to main branch?

akira:
Hello- thanks for the fix! Any thoughts on merging this? Would be great to have for my use cases! 🙇

mernmic:
Fix while waiting for PR merge.

Initialize ArqRedis with a connection pool including your retry helper.

```
class WorkerSettingsBase:
    redis_pool = ArqRedis(
        connection_pool=redis.asyncio.ConnectionPool.from_url(
            url=RedisSecrets.REDIS_URL,
            retry_on_timeout=True,
            retry_on_error=[
                redis.asyncio.TimeoutError,
                redis.asyncio.ConnectionError,
            ],
            retry=Retry(
                backoff=CustomRedisBackoff(),
                retries=5,
            )
        )
    ),
   ....

if __name__ == "__main__":
    arq.run_worker(WorkerSettingsBase)
```

redis_pool parameter can be found in docs here:
https://arq-docs.helpmanual.io/#arq.worker.Worker

Example custom backoff class, we are using for more granular access and logging but you can use any redis retry helper
```
class CustomRedisBackoff(AbstractBackoff):
    def __init__(self, cap=0.512, base=0.008):
        """
        `cap`: maximum backoff time in seconds
        `base`: base backoff time in seconds
        """
        self.backoff = ExponentialBackoff(cap=cap, base=base)

    def compute(self, failures):
        sleep = self.backoff.compute(failures)
        logger.info(
            {
                "message": "redis connection error, triggering backoff",
                "failures": failures,
                "backoff": repr(self.backoff),
                "sleep (secs)": sleep,
            }
        )
        return sleep
```

samuelcolvin (approved):
Thanks so much for this, sorry for the long delay.

samuelcolvin:
@JonasKs @Kludex we might be able to use redis's retry logic to remove our retry stuff - https://github.com/samuelcolvin/arq/blob/94cd8782b4f0764a17962186a349d32125cb98e3/arq/connections.py#L258-L286

samuelcolvin:

PR #392, opened 2023-03-18
Fix connections.py: allow to connect to Redis using a Unix socket URL…

drygdryg:
Allow to connect to Redis using a Unix socket URL without specifying the database number using querystring parameter.
Currently, when trying to connect via a Unix socket, specifying a URL without a db number, we get a parsing error. This fix allows to skip specifying the db number and set `db=0` by default .
This code demonstrates buggy behavior of the URL parser:
```python

settings = RedisSettings.from_dsn('unix:///run/redis/redis.sock')
```
Causes error:

```
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    settings = RedisSettings.from_dsn('unix:///run/redis/redis.sock')
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "venv/lib/python3.11/site-packages/arq/connections.py", line 59, in from_dsn
    database = int(conf.path.lstrip('/')) if conf.path else 0
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'run/redis/redis.sock'
```
While this works correcrly:
```python

settings = RedisSettings.from_dsn('unix:///run/redis/redis.sock?db=0')
```


JonasKs:
Thanks for the PR 😊 Would love a test so we ensure it won't happen again.

samuelcolvin:
Looks good, I agree on tests.

samuelcolvin:

PR #396, opened 2023-05-10
Allow infinite retry

vvmruder:
In some contexts (cluster) it is useful to allow infinite retries `float('inf')`. Especially in clusters redis might be not available for some time but we don't want the worker to crash on that. Instead it just should wait until redis comes available again.

`float('inf')` always compare greater than any float or int. So it should be safe to use. But `str(float('inf'))` translates to `'inf'` and that's why this change is necessary.


samuelcolvin:
thanks so much.

samuelcolvin:

PR #398, opened 2023-05-20
Delete setup.py

stradivari96:
Fixed here:
https://github.blog/changelog/2023-02-13-dependency-graph-supports-the-python-pep-621-standard/

And it looks like it is working fine in pydantic
https://github.com/pydantic/pydantic/pull/4881


samuelcolvin:
Thank you.

PR #399, opened 2023-05-20
deps: Upgrade redis stubs

sondrelg:
Quick clean-up PR. Noticed there were a lot of Redis-pipeline `mypy` errors that shouldn't occur with the latest version of types-redis.

Not 100% sure if `Pipeline[bytes]` is the appropriate annotation here, so please let me know if not. Also, if you have a better suggestion for how to fix types inference for the pipeline variables, that's of course very welcome; the manual definitions are not ideal.


JonasKs (approved):
Nice! 😊

phy1729:
The pipeline types aren't being inferred correctly due to mypy. If the latest mypy version is used, those hints are no longer needed.

I've submitted #403 to update mypy, but it'll conflict with your PR. If this PR is merged first I can rebase and add a commit to remove the hints.

samuelcolvin:
Thanks so much for the contribution, this is replaced by https://github.com/samuelcolvin/arq/pull/439 where I've fixed lots of related things.

PR #403, opened 2023-05-26
Update mypy

phy1729:
Mypy changed the default for the [implicit_optional setting](https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_optional), so I included that in this PR as well. Error codes are [shown by default](https://mypy.readthedocs.io/en/stable/config_file.html#confval-hide_error_codes) now and the `ollow_imports = 'silent'` setting is no longer required for mypy to pass.


samuelcolvin:
Thanks so much for the contribution, this is replaced by https://github.com/samuelcolvin/arq/pull/439 where I've fixed lots of related things.

Issue #405, opened 2023-05-31
Job aborts failing


rm-21:
I have setup a sample project: [arq-scale-test](https://github.com/rm-21/arq-scale-test). I love Arq but somehow task abort is failing for me. Any help or insights would be appreciated.

I have a simple long running tasks that sleeps for 60 seconds. 
5 tasks are queued and started in `app.py`.
In `checks.py`, I attempt to check the status of the task and stop it. I am successfully able to retrieve the task status. But while calling `await job.abort()`, it just hangs. The task actually gets completed within 60 seconds & then I get a job aborted as `False`. 

Adding the code below as well:

```python
# config.py
from typing import Any

from arq.connections import RedisSettings
from pydantic import BaseSettings
from pydantic import validator


class Settings(BaseSettings):
    ENVIRONMENT: str | None

    # Redis backend
    REDIS_HOST: str
    REDIS_PORT: str

    # ARQ Config
    ARQ_REDIS_SETTINGS: RedisSettings | None = None
    ARQ_MAX_JOBS: int = 5
    ARQ_JOB_TIMEOUT: int = 9 * 60 * 60  # 9hrs in seconds
    ARQ_KEEP_RESULT: int = 9 * 60 * 60  # 9hrs in seconds
    ARQ_MAX_TRIES: int = 1
    ARQ_HEALTH_CHECK_INTERVAL: int = 5 * 60  # 5 minutes in seconds
    ARQ_RETRY_JOBS: bool = False
    ARQ_ALLOW_ABORT_JOBS: bool = True

    @validator("ARQ_REDIS_SETTINGS")
    @classmethod
    def assemble_arq_redis_settings(
        cls, value: RedisSettings | None, values: dict[str, Any]
    ) -> RedisSettings:
        if isinstance(value, RedisSettings):
            return value
        settings = RedisSettings(
            host=values["REDIS_HOST"],
            port=values["REDIS_PORT"],
            database=1,
        )
        return settings

    class Config(BaseSettings.Config):
        case_sensitive: bool = True
        env_file: str = ".env"


def get_settings() -> Settings:
    return Settings()


settings: Settings = get_settings()
```

```python
# arq_app.py
import asyncio
from typing import Any

from .config import settings


async def long_running_func(
    ctx: dict[Any, Any] | None,
) -> None:
    await asyncio.sleep(60)


def worker_dispatcher():
    class ArqWorker:
        # Functions to schedule
        functions = [long_running_func]

        # ARQ settings
[...]

PR #408, opened 2023-06-10
🔨 Adding a job counter to address Semaphore issues

rm-21:
**Issue** #405  
Having a semaphore in the same event loop, while having the max number of jobs running, blocks heartbeat & cancellation tasks.

**Possible solution**
* Keep the semaphore value to `max_jobs + 1`
* Have a `job_counter` that increments when the semaphore is acquired and decrements when the semaphore is released
* If the `job_counter == max_jobs`, release the semaphore immediately and return from the `start_jobs` function, basically foregoing starting new jobs.

The solution seems to work. The counter is threadsafe because we always increment or decrement it **before** releasing the semaphore. Quick test app below:
* Define the worker definition in `arq_app.py`
* Start 4 workers in different processes from `start_workers.py`
* Schedule more jobs than max for each worker from `arq_app.py`
* Check `worker.log` for the first heartbeat happening
* Cancel a task from `cancellations.py`
* Check `worker.log` for the latest heartbeat + cancelleation + new job

```python
# arq_app.py
import asyncio
import datetime as dt
from typing import Any

from arq import create_pool

REDIS = RedisSettings(host="127.0.0.1", port=6379, database=1)

async def long_task(ctx: dict[Any, Any], num:int, sleeper: float):
    print(f"[{dt.datetime.now()}] {num}: Task entered & started...")

    # Mock call
    await asyncio.sleep(1)
    
    # Fail randomly
    # import random
    # if random.random() > 0.5:
    #     raise RuntimeError(f"[{dt.datetime.now()}] {num}: Task randomly failed...")
    
    # Balance task
    await asyncio.sleep(sleeper)

    print(f"[{dt.datetime.now()}] {num}: Task finished...")



def worker_dispatcher(worker_queue_name: str):
    class ArqWorker:
        # Functions to schedule
        functions = [long_task]

        # ARQ settings
        redis_settings = REDIS
        max_jobs = 5
        job_timeout = 9 * 60 * 60 # 9hrs
        keep_result = 9 * 60 * 60 # 9hrs
        health_check_interval = 5
        retry_jobs = False
        allow_abort_jobs = True

        queue_name = worker_queue_name

    return ArqWorker


async def main():
    redis = await create_pool(settings_=REDIS)

    for idx in range(32):
        await redis.enqueue_job("long_task", idx, 6 * 60 * 60, _queue_name=f"queue{idx % 4}")


if __name__ == "__main__":
    asyncio.run(main())
``` 

```python
# start_workers.py
import multiprocessing as mp
import logging

from arq import run_worker
from .arq_app import worker_dispatcher

def start_worker(queue_name: str):
    logging.basicConfig(filename='worker.log', level=logging.INFO,
                        format='%(asctime)s [%(process)d] %(levelname)s: %(message)s')
    logging.info(f'Starting worker for {queue_name}')
    run_worker(worker_dispatcher(queue_name))

if __name__ == "__main__":
    for i in range(4):
        mp.Process(target=start_worker, args=(f"queue{i}",)).start()
```

```python
# cancellation.py
import asyncio

from arq import create_pool
from arq.jobs import Job

REDIS = RedisSettings(host="127.0.0.1", port=6379, database=1)

async def job_status(task_id: str, stop: bool = False) -> None:
    job: Job = Job(
        job_id=task_id,
        redis=await create_pool(settings_=REDIS),
    )

    print(f"ID {task_id}: {await job.status()}")

    if stop:
        result: bool = await job.abort()
        print(f"ID {task_id}: Aborted: {result}")


async def main() -> None:
    await job_status(task_id="258a1b3ec3a84913a42d992b970841da", stop=True)


if __name__ == "__main__":
    asyncio.run(main())
```

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -236,7 +236,11 @@ class Worker:
         self.on_job_start = on_job_start
         self.on_job_end = on_job_end
         self.after_job_end = after_job_end
-        self.sem = asyncio.BoundedSemaphore(max_jobs)
+
+        self.max_jobs = max_jobs
+        self.sem = asyncio.BoundedSemaphore(max_jobs + 1)
+        self.job_counter: int = 0
+
         self.job_timeout_s = to_seconds(job_timeout)
         self.keep_result_s = to_seconds(keep_result)
         self.keep_result_forever = keep_result_forever
@@ -374,13 +378,13 @@ class Worker:
                 return
             count = min(burst_jobs_remaining, count)
         if self.allow_pick_jobs:
-            async with self.sem:  # don't bother with zrangebyscore until we have "space" to run the jobs
+            if self.job_counter < self.max_jobs:
                 now = timestamp_ms()
                 job_ids = await self.pool.zrangebyscore(
                     self.queue_name, min=float('-inf'), start=self._queue_read_offset, num=count, max=now
                 )
 
-            await self.start_jobs(job_ids)
+                await self.start_jobs(job_ids)
 
         if self.allow_abort_jobs:
             await self._cancel_aborted_jobs()
@@ -419,12 +423,23 @@ class Worker:
             self.aborting_tasks.update(aborted)
             await self.pool.zrem(abort_jobs_ss, *aborted)
 
+    def _release_sem_dec_counter_on_complete(self) -> None:
+        self.job_counter = self.job_counter - 1
+        self.sem.release()
+
     async def start_jobs(self, job_ids: List[bytes]) -> None:
         """
         For each job id, get the job definition, check it's not running and start it in a task
         """
         for job_id_b in job_ids:
             await self.sem.acquire()
+
+            if self.job_counter >= self.max_jobs:
+                self.sem.release()
+                return None
+
+            self.job_counter = self.job_counter + 1
+
             job_id = job_id_b.decode()
             in_progress_key = in_progress_key_prefix + job_id
             async with self.pool.pipeline(transaction=True) as pipe:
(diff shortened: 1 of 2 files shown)


JonasKs:
Thank you! 😊

Would be great if you could add a test that ensure we don't regress in the future. 😊

rm-21:
Could you suggest a test? I couldn't come up with an appropriate one.

JonasKs:
Set max jobs to 1, queue a task and ensure health check is still logged? This should fail without this implementation, right?

I'm not at home, but I'm pretty sure there's a health check test already implemented.

rm-21:
I have setup a test that aims to cancel the job when max_jobs queue is full. The test is failing for Python 3.7. The logs show that the job was enqueued and cancelled. There is a stack trace for ConnectionError from Redis in between the 2 log lines. Can you help me debug whenever you can make some time? 

I do think just re-running the CI job for 3.7 should resolve this.

arq/worker.py
```diff
@@ -419,12 +423,23 @@ async def _cancel_aborted_jobs(self) -> None:
...
     async def start_jobs(self, job_ids: List[bytes]) -> None:
         """
         For each job id, get the job definition, check it's not running and start it in a task
         """
         for job_id_b in job_ids:
             await self.sem.acquire()
+
+            if self.job_counter >= self.max_jobs:
+                self.sem.release()
+                return None
```
JonasKs, line 439: Just `return` here.

JonasKs (approved):
Thank you 😊   
This looks good to me. 

Samuel is busy with v2 of Pydantic, so I wouldn't expect this to be merged until that happens, but you can always use your own fork in the mean time 😊

samuelcolvin (approved):
Sorry the slow reply.

From a brief review, LGTM.

@JonasKs merge if you're happy.

rm-21:
Hey @JonasKs! Any update on when can this be merged?

samuelcolvin:

PR #409, opened 2023-07-05
Improves after_job_end hook

AngellusMortis:
Since `finish_job` can fail, I adjusted the `after_job_end` hook to always receive a copy of the the `JobResult` that would get written into redis.

I also updated all of the other places that recorded a failing job to call `after_job_end` as well and added a `asyncio.shield` to `after_job_end` as it is a great way to do something with the job result (like generate metrics for New Relic/Prometheus/etc.).

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -624,6 +624,7 @@ class Worker:
         )
         result_timeout_s = self.keep_result_s if function.keep_result_s is None else function.keep_result_s
         result_data = None
+        job_result = None
         if result is not no_result and (keep_result_forever or result_timeout_s > 0):
             job_result, result_data = _create_result(
                 function_name,
@@ -687,9 +688,10 @@ class Worker:
                 tr.delete(*delete_keys)  # type: ignore[unused-coroutine]
             await tr.execute()
 
-    async def _after_job_end(self, ctx: Dict[Any, Any], job_result: JobResult) -> None:
+    async def _after_job_end(self, ctx: Dict[Any, Any], job_result: Optional[JobResult]) -> None:
         if self.after_job_end:
-            ctx['job_result'] = job_result
+            if job_result:
+                ctx['job_result'] = job_result
             await self.after_job_end(ctx)
 
     async def finish_failed_job(self, job_id: str, result_data: Optional[bytes]) -> None:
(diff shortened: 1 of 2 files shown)


gerazenobi:
@AngellusMortis hi there 👋 

I am taking advantage of your deeper knowledge regarding this hook.

**question**: without this PR the hook `after_job_end` only works when there were no errors and jobs successfully finished?
E.g. if the job raised an exception, the hook won't be called ?

If that is the case, would there be a work around using other hooks ?

Thanks in advance

arq/worker.py
```diff
@@ -486,9 +486,10 @@ async def run_job(self, job_id: str, score: int) -> None:  # noqa: C901
         args: Tuple[Any, ...] = ()
         kwargs: Dict[Any, Any] = {}
 
-        async def job_failed(exc: BaseException) -> None:
+        async def job_failed(exc: BaseException, ref: Optional[str] = None) -> None:
+            ref = ref or f'{job_id}:{function_name}'
```
samuelcolvin, line 490:
```suggestion
        async def job_failed(exc: BaseException, ref: str = f'{job_id}:{function_name}') -> None:
```

arq/worker.py
```diff
@@ -701,6 +687,12 @@ async def finish_job(
                 tr.delete(*delete_keys)  # type: ignore[unused-coroutine]
             await tr.execute()
 
+    async def _after_job_end(self, ctx: Dict[Any, Any], job_result: Optional[JobResult]) -> None:
```
samuelcolvin, line 690: please add a docstring.

arq/worker.py
```diff
@@ -886,6 +878,50 @@ def run_worker(settings_cls: 'WorkerSettingsType', **kwargs: Any) -> Worker:
...
+    enqueue_time_ms: int,
+    success: bool,
+    result: Any,
+    start_ms: int,
+    finished_ms: int,
+    ref: str,
+    queue_name: str,
+    *,
+    serializer: Optional[Serializer] = None,
+) -> Tuple[JobResult, Optional[bytes]]:
```
samuelcolvin, line 895: please add a docstring explaining what this does.

samuelcolvin (commented):
In general looks good, mostly just needs some docstrings.

AngellusMortis:
I no longer care.

PR #427, opened 2024-02-13
Add python 3.12 for test

brian-goo:


samuelcolvin:
Thanks so much for the contribution, this is replaced by #439 where I've fixed lots of related things.

Issue #417, opened 2023-10-22
Function level keep_result_s not respected for errors


SoftMemes:
In `finish_failed_job`, there's only a check on self.keep_result_s, not the value configured from the function. As a result, any failed tasks are stored for potentially a very long time and cannot be re-enqueued with the same job id, even if configured on the function level to not store a result.

This also relates to https://github.com/samuelcolvin/arq/issues/416 as currently it's impossible to follow this pattern for tasks that can fail, e.g. trying to re-enqueue a task that you want to have eventually succeed does nothing of the worker is configured to keep values.

Issue #416, opened 2023-10-12
Question: How can one self-re-enqueue a job when using _job_id?


joshwilson-dbx:
I have a use case where once a job completes, i would like it to continuously re-schedule itself for another run at some point in the future. I would also like to ensure that only one instance is queued/running at any given time, so I'm using the `_job_id` parameter when enqueuing. I cannot use the `cron` functionality as the delay time is somewhat dynamic and not easily translated to cron. 

Options that I've explored so far:
1) Simply call `redis.enqueue_job(..., _job_id=ctx['job_id'])` from within the job itself
    - This doesn't work since the current job is still active and prevents the new enqueuing
2) Raise a `Retry` exception from within the job after the work has completed
    - This seems like an abuse of this functionality
    - Will likely run into trouble with the `_expires` and `max_tries` settings
3) Set `keep_result=0` and enqueue a second job with a small delay that in turn re-enqueues the original job again
    - Works but is cumbersome and may introduce a race condition
    - Needs a second job function just to enqueue the primary job again
4) Use `keep_result=0` and re-enqueue in the `after_job_end` function to be sure the job and result keys are no longer present so the re-enqueue can occur.
    - Will probably need to dedicate a specific queue and workers for these jobs
    - Risks the programmer error of enqueuing in the wrong queue

Is there a better way to do this?

gerazenobi:
Hi @joshwilson-dbx, I think we have somewhat similar use cases, thought I'd reference my issue here https://github.com/samuelcolvin/arq/issues/394 to _vote up_ this scenario and also in case it helps to read about the problems I had encountered with aborting/deleting.

joshwilson-dbx:
Thanks for taking a look @gerazenobi. I agree our issues both arise from not being able to manage redundant job behavior well with the current API. I don't know what the solution would be yet, but perhaps we could add additional parameters to `enqueue_job()` to specify how to manage these collisions.

It doesn't seem like this kind of configuration would belong on the worker side of things. Maybe it belongs on the `Job` class?

I also wonder if there's a need to formalize the concept of a Queue in the code. Right now it looks like it's just a convention specified by passing around a queue name between the various job and worker functionality.

gerazenobi:
@joshwilson-dbx 
[...]

PR #433, opened 2024-02-20
Use function's keep_results configuration when storing failed jobs results

joshwilson-dbx:
Taking a stab at using a function's (or cron job's) configuration for keeping the job results when the job fails.

I don't think this is a complete solution yet, because there are a few cases where the job results are written to redis using the Worker's configuration because we can't get the Function in order to use it's config.

But maybe this is a situation where it's good enough to improve the situation for the most common situation?

Addresses #417
Relates to #416

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -486,7 +486,7 @@ class Worker:
         args: Tuple[Any, ...] = ()
         kwargs: Dict[Any, Any] = {}
 
-        async def job_failed(exc: BaseException) -> None:
+        async def job_failed(exc: BaseException, function: Optional[Union[Function, CronJob]]) -> None:
             self.jobs_failed += 1
             result_data_ = serialize_result(
                 function=function_name,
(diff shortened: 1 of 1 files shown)


arq/worker.py
```diff
@@ -483,6 +483,7 @@ async def run_job(self, job_id: str, score: int) -> None:  # noqa: C901
                 abort_job = False
 
         function_name, enqueue_time_ms = '<unknown>', 0
+        function: Optional[Union[Function, CronJob]] = None
```
samuelcolvin, line 486:
```suggestion
        function: Union[Function, CronJob, None] = None
```

samuelcolvin (commented):
Otherwise I think this looks okay, @JonasKs WDYT?

JonasKs:
Agree, I think this looks good. 😊

epicwhale:
Thanks for looking at this one! Any chance this can enter the v0.26.0b1 release? :)

Issue #437, opened 2024-03-17
Future plan for Arq


samuelcolvin:
Arq was the first real open source project I ever created, back in 2016. That was long before Pydantic, FastAPI, ParamSpec, or even Redis Streams.

I remember a sense of incredulity that I couldn't find an async variant of [rq](https://github.com/rq) (which I was helping to maintain at the time), surely I wasn't the only person wanting to queue jobs in async code? Apparently at the time I was.

Fast forward eight years, and I'm definitely not the only person try to queue jobs in an async world.

Hence my incredulity has only grown - there's still no ubiquitous queuing library for async Python, and despite neglect, Arq still seems to work well for lots of people, I've used it in every role I've had since, and for the most part it just works.

---

That said, Arq needs some love, and since we're now using it at Pydantic, I think we should have the resources to provide that love later this year. This is a rough plan of what I propose to do. 

Feedback very welcome from all, but especially @JonasKs and @pydantic-maintainers (who apparently I can't tag here :-()

In summary I want to significantly refactor the internals, and update the API in a backwards compatible way.

### 1. `ParamSpec` and type safety

The most important change we should make is to make Arq typesafe using [`ParamSpec` and `Concatenate`](https://docs.python.org/3/library/typing.html#typing.ParamSpec), I have a partially working demonstration of how this will work below.

We'll provide the same or similar method for enqueuing a job via it's name for cases where the worker code is not accessible where jobs are enqueued.

We should be able to do this while still supporting the current API to make migration as easy as possible.

### 2. Redis Streams

The second most important change is to adopt [Redis streams](https://redis.io/docs/data-types/streams/) which read like they were designed for exactly this application, they mean we can effectively guarantee only one execution while still being resilient to unexpected shutdown (Jobs shutdown during execution will be rerun later).

This should be possible without breaking the current API at all.

### 3. Avoid sorted set for immediate jobs

Current Arq is slower than it should be because it uses a [sorted set](https://redis.io/docs/data-types/sorted-sets/) to queue all jobs, the sorted set provides two things:
* a way to avoid loosing jobs that are cancelled during shutdown or fail - this should be solved with streams (see above)
* a way to defer jobs so they're run in the future

[...]

PR #439, opened 2024-04-01
3.12 support, drop 3.7, uprev dependencies

samuelcolvin:
Related to #437.

First step is to get a release out with all the pending PRs, and 3.12 support.


tests/test_jobs.py
```diff
@@ -110,17 +110,6 @@ async def foobar(ctx, *args, **kwargs):
     ]
 
 
-async def test_enqueue_job_with_unix_socket(worker, unix_socket_path):
```
samuelcolvin, line 113: redislite doesn't work with python 3.12, and getting a socket connection wasn't simple otherwise, hence removing the fixture and this test.

PR #446, opened 2024-04-19
Fix "retry_on_error" type annotation

armicron:
When a list of classes (not instances) is assigned to `RedisSettings.retry_on_error`, mypy finds an error:
```
$ make mypy
mypy arq
arq/poc_annotations.py:5: error: List item 0 has incompatible type "type[ConnectionError]"; expected "Exception"  [list-item]
Found 1 error in 1 file (checked 13 source files)
make: *** [Makefile:46: mypy] Error 1
```
Redis library for the similar parameter has the type annotation `Optional[List[Type[Exception]]]` ([source](https://github.com/redis/redis-py/blob/07fc339b4a4088c1ff052527685ebdde43dfc4be/redis/asyncio/cluster.py#L266))

How to reproduce:
create a file `arq/poc_annotations.py`

```python
from redis.exceptions import ConnectionError


settings = RedisSettings(retry_on_error=[ConnectionError])
```

### DIFF
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -4,7 +4,7 @@ import logging
 from dataclasses import dataclass
 from datetime import datetime, timedelta
 from operator import attrgetter
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, Callable, List, Optional, Tuple, Type, Union, cast
 from urllib.parse import parse_qs, urlparse
 from uuid import uuid4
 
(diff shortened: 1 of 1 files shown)


ivorpeles:
Ran into this as well, would really like to see this merged. Can further confirm that if you pass in instances mypy is satisfied but my code breaks.

Even in the test code for arq the correct types are passed, even though this isn't detected by CI:
https://github.com/samuelcolvin/arq/blob/1315583f170512fc63622af0200a2855e057a58d/tests/conftest.py#L58

PR #448, opened 2024-04-25
Set the result after cleaning the queue to reduce stuck transaction

isra17:
We are seeing spikes of `ResultNotFound` error from time to time. We are unable to reproduce reliably, but we think it might be coming from ARQ/Redis. Our asyncio tasks are not blocked and not every job get `ResultNotFound` during the spikes.

So my current theory is that we have lock contention in Redis in the `finish_job` function. It could be possible that deleting the `job_id` from the queue is blocked for more than the result TTL (Which we have at 10s) then by the time the pipeline execute, the result key with already be expired.

This change simply move the result key set after we delete the `job_id` from the busy queue that might block on a lock. If my theory is right, we should stop seeing the `ResultNotFound` error.

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -691,12 +691,12 @@ class Worker:
                 tr.pexpire(in_progress_key, to_ms(keep_in_progress))
 
             if finish:
-                if result_data:
-                    expire = None if keep_result_forever else result_timeout_s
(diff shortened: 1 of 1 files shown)


isra17:
Nevermind, this doesn't fix it, I'm going to research a bit more into redis transaction model first ;)

PR #451, opened 2024-05-04
Add Redis Streams option for job delivery

ajac-zero:
This pull request adds a basic implementation of Redis Streams, in order to avoid polling for new jobs in the worker and reduce latency, in accordance with objective 4 of issue #437. 

To create a worker that listens to a Redis Stream, we can use the cli or specify it in the code directly.

**CLI:**
`arq worker.WorkerSettings --stream`

**Code:**
```python
class WorkerSettings:
    functions = [...]
    stream = True
    ...
```

On the client, they must specify that they want to deliver a job to a worker through a Redis Stream.
```python
await redis.enqueue_job('hello_world', _use_stream=True)
```

Here are the results of a _very_ simple benchmark that showcases the potential of using Redis Streams for improved latency.

**Polling:**
Captura de pantalla 2024-05-04 a la(s) 8 12 28 a m
Average time: 0.268s

**Streaming:**
Captura de pantalla 2024-05-04 a la(s) 8 15 34 a m
Average time: 0.012s

### DIFF
--- a/arq/cli.py
+++ b/arq/cli.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
     from .typing import WorkerSettingsType
 
 burst_help = 'Batch mode: exit once no jobs are found in any queue.'
+stream_help = 'Stream mode: use redis streams for job delivery. Does not support batch mode.'
 health_check_help = 'Health Check: run a health check and exit.'
 watch_help = 'Watch a directory and reload the worker upon changes.'
 verbose_help = 'Enable verbose output.'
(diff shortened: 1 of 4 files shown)


gaby:
@ajac-zero  This may need a unit-test with `stream` enabled.

ajac-zero:
Sure thing @gaby . I was wondering how I should go about that...

What I started doing was add a stream parameter to the worker tests, and then wrap them so they run twice, once with stream and once without.

But I feel this might not be the best way to do things, maybe I should focus on some vital tests? What do you suggest?

gaby:
@ajac-zero That's probably a great starting point, running current tests with "stream" set to false. Then running the test suite with "stream" set to True. This will require setting the source of the data to use Streams.

ajac-zero:
@gaby I finally got around to writing the unit tests. I added a new `stream_worker` pytest fixture and used pytest parametrize to basically run all the worker test suite twice, first with polling and then with streaming. All passing 😸.

samuelcolvin:
This looks like great work, but I wonder if it will add complexity to the migration/rewrite required by #437.

Given the size of the change required by #437, I think I'll work on a clean-room rewrite, then add compatibility shims for the existing methods — this might make that work more complicated.

ajac-zero:
@samuelcolvin Thanks! Yes, it might be better to wait and build on top of the new version. 

I'm really looking forward to the new changes. The DAG and type safety sound awesome. Lmk if I can contribute in some way.

sandeep2rawat:
Hi folks, thanks for the implementation, really a needful change.
Any update on this feat ? And anyone tested it with multiple workers
Seems there's a similar MR https://github.com/python-arq/arq/pull/492.

Is there any plan to merge this feat ?

PR #462, opened 2024-05-27
proof of concept UI, created using FastUI to display status of the jobs in the browser

tobiasz-gleba:
Hi, I've created a proof of concept UI, using FastUI to display the status of the jobs in the browser. If you @samuelcolvin can take a look 👀, and tell me if it is worth it to develop it and include it in the package 🥶

Screenshot 2024-05-27 at 23 43 04
Screenshot 2024-05-27 at 23 43 11
Screenshot 2024-05-27 at 23 43 16

It could be mounted as a sub-application in the fastApi ecosystem, sth like https://www.hangfire.io/ in .NET

What do you think? Maybe I should create another package, just for the UI to not mess arq up with separate functionality.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,55 @@
+{
+    "name": "CentralAPI",
+    "image": "python:3.11",
+    "workspaceFolder": "/app",
+    "remoteUser": "root",
+    "overrideCommand": true,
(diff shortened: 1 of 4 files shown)


JonasKs:
That looks awesome. 

Whether it should be in a separate package or not I'll let Samuel comment on, but I support this work!!

samuelcolvin:
Thanks so much @tobiasz-gleba, looks very useful.

As you suggested, I think this should be in a separate repo.

PR #463, opened 2024-06-07
Stringify arguments and results only when logging is enabled

isra17:
One of our profiling profile shows that 40% of computing time is spent in `repr` call from ARQ. At the moment, even without logging this get called, which can be expensive on large inputs.


samuelcolvin (commented):
LGTM, can you add a test.

isra17:
> LGTM, can you add a test.

Done

PR #470, opened 2024-07-18
Uses `testcontainers` to provide a redis instance for the unit tests

chrisguidry:
We can't assume that a dev would want the unit tests to connect to the locally
running redis at `localhost:6379`, so I'd rather shift that to assuming that the
user has a Docker-compatible system available.  This makes the test suite easier
to reason about and lets us control the redis version with more precision, and
it makes the initial startup experience easier for new devs.


.github/workflows/ci.yml
```diff
@@ -5,96 +5,90 @@ on:
     branches:
       - main
     tags:
-      - '**'
+      - "**"
```
chrisguidry, line 8: Sorry, y'all, my VSCode has strong feelings about YAML files

.github/workflows/ci.yml
```diff
@@ -5,96 +5,90 @@ on:
...
+      - run: pre-commit run -a --verbose
+        env:
+          SKIP: no-commit-to-branch
 
   docs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v4
+      - uses: actions/checkout@v4
```
samuelcolvin, line 33: does this need changing now (or at all)? It makes the other changes hear hard to see, and add unnecessary churn.

samuelcolvin (commented):
Everything else looks good to me.

.github/workflows/ci.yml
chrisguidry: Yep sorry about that, I'll get this this tuned back to just the salient changes

samuelcolvin (approved):

samuelcolvin:
LGTM, merge if you're happy.

JonasKs (approved):

PR #487, opened 2024-11-26
Fix race condition on task retry

RB387:
Related issue:
https://github.com/python-arq/arq/issues/482


brylie-wolt (approved):

RB387:
@samuelcolvin @chrisguidry 🙏

PR #488, opened 2024-11-30
Bumps version for next release

chrisguidry:
This bumps the version to 0.26.3 to prep for the next release


chrisguidry:
@samuelcolvin I forgot that I needed to bump the version prior to the release, so this will queue up for the _next_ one

PR #496, opened 2025-01-28
add workers arg to CLI

Graeme22:
## Description

This PR adds a new integer argument, `--workers` (or `-w`), to the CLI. This will cause N-1 additional worker processes to be spawned with the same configuration. This should be useful for the common case where more than one worker per machine/container is desired.

## Usage

```bash
arq --workers 4 example.WorkerSettings
```

This also works with the `--watch` flag: all workers will be restarted upon the detection of changes.


Graeme22:
Not sure why this isn't working on Python 3.12/Redis 5 consistently!

PR #504, opened 2025-07-10
Feat(Stream): Use redis stream

saber-solooki:
The whole PR got inspired from these two PR: https://github.com/python-arq/arq/pull/451 and https://github.com/python-arq/arq/pull/492 (Great job). Also I took a look at this [library](https://github.com/tastyware/streaq) to get some idea. (Great Library)

I decided to create another PR for redis stream because I noticed some improvements on those PR which I try to explain here:

https://github.com/python-arq/arq/pull/451 try to use redis stream but still using [_pool_iteration](https://github.com/python-arq/arq/pull/451/files#diff-6bc16b0b2608a9226aeb5a37f083c6a49ca6c15f1ec457eb7fac3327e2f93c4aR396) which is somehow utilizing the old implementation without any benefit. Not sure what stream feature is actually happening there.
However, it has some nice ideas, like passing the flag to decide whether the user wants to use the stream or not.

On the other hand, the second PR has a more accurate usage of stream, but still has some ideas for implementation which I didn't understand correctly. On the other hand, it assume the user only wants to use stream by [removing the old](https://github.com/python-arq/arq/pull/492/files#diff-27b77fc91f229c8075c3a1fe84186a058cc661154a28b56f2bed8d3ec28549edL175) implementation while enqueuing the job.

The second thing that I can't support was [registering 3 different task](https://github.com/python-arq/arq/pull/492/files#diff-6bc16b0b2608a9226aeb5a37f083c6a49ca6c15f1ec457eb7fac3327e2f93c4aR388-R390) in the worker:
```
asyncio.ensure_future(self.run_delayed_queue_poller()),
asyncio.ensure_future(self.run_stream_reader()),
asyncio.ensure_future(self.run_idle_consumer_cleanup()),
```

The run_delayed_queue_poller always tries to read jobs from the simple queue and put it in the stream. If you take a close look, it has some logic, some places to keep putting the task in the queue, and again read it here and put it in the stream. 

The run_stream_reader is the actual method to read tasks from the stream.
The run_idle_consumer_cleanup also tries to remove idle consumers from the Redis. Imagine if we have multiple workers who always try to clean Redis. It can cause some clean operations on Redis, which I believe we can improve. I try to address removing the consumer in the close method like this [part of the library](https://github.com/tastyware/streaq/blob/v3.0.0/streaq/worker.py#L1087-L1091). We need to clean up the consumer because we are generating a consumer name each time the worker is brought up. I can't imagine any easy way to consistently generate a unique name for our consumers.

Also I removed [_unclaim_job](https://github.com/python-arq/arq/pull/492/files#diff-6bc16b0b2608a9226aeb5a37f083c6a49ca6c15f1ec457eb7fac3327e2f93c4aR625) from the implementation since from my understanding, the usage was to unclaim any job that consumer is trying to get while it is already taken by another consumer. Well, that is the whole point of using Redis stream to prevent such scenario and it shouldn't happen unless there is something wrong with Redis.

As a final note, I also added some improvements in some places to make the logic more robust, like if self.use_stream

It is worth mentioning that we have been using this implementation on our product environment for almost 2 months and everything is working as expected. All the benefits and all metric improvements are almost close to what is mentioned in [this PR](https://github.com/python-arq/arq/pull/492). 

Thanks again @RB387 and @ajac-zero for your great job. If you are still interested in the topic, I appreciate having you review 🙏 

If everything is fine from the owner, I will add the final tests to have 100% test coverage regarding my changes.

### DIFF
--- a/arq/connections.py
+++ b/arq/connections.py
@@ -13,8 +13,16 @@ from redis.asyncio.retry import Retry
 from redis.asyncio.sentinel import Sentinel
 from redis.exceptions import RedisError, WatchError
 
-from .constants import default_queue_name, expires_extra_ms, job_key_prefix, result_key_prefix
+from .constants import (
+    default_queue_name,
+    expires_extra_ms,
+    job_key_prefix,
+    job_message_id_prefix,
+    result_key_prefix,
+    stream_key_suffix,
+)
 from .jobs import Deserializer, Job, JobDef, JobResult, Serializer, deserialize_job, serialize_job
+from .lua_script import publish_job_lua
 from .utils import timestamp_ms, to_ms, to_unix_ms
 
 logger = logging.getLogger('arq.connections')
@@ -114,6 +122,7 @@ class ArqRedis(BaseRedis):
         if pool_or_conn:
             kwargs['connection_pool'] = pool_or_conn
         self.expires_extra_ms = expires_extra_ms
+        self.publish_job_sha = None
         super().__init__(**kwargs)
 
     async def enqueue_job(
@@ -126,6 +135,7 @@ class ArqRedis(BaseRedis):
         _defer_by: Union[None, int, float, timedelta] = None,
         _expires: Union[None, int, float, timedelta] = None,
         _job_try: Optional[int] = None,
+        _use_stream: bool = False,
         **kwargs: Any,
     ) -> Optional[Job]:
         """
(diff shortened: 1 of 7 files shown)


Aduomas (approved):
looks good m8

kmariunas:
solid work boss, would've really helped us out 30 minutes ago..

PR #509, opened 2025-10-14
Add python3.14 version support

a3lme:
This PR handles the following [asyncio behavioral changes](https://docs.python.org/3.14/whatsnew/3.14.html#id11) introduced in python3.14:

[asyncio.get_event_loop()](https://docs.python.org/3.14/library/asyncio-eventloop.html#asyncio.get_event_loop) now raises a RuntimeError if there is no current event loop, and no longer implicitly creates an event loop.

To support 3.14, this PR:
1. Fixes [Python] RuntimeError: no running event loop by checking version of python and use asyncio.new_event_loop()
2. Update dependencies versions for no conflicts with 3.14

### DIFF
--- a/arq/worker.py
+++ b/arq/worker.py
@@ -2,6 +2,7 @@ import asyncio
 import contextlib
 import inspect
 import logging
+import sys
 import signal
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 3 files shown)


TomFaulkner:
Thanks for this, @a3lme.  Easy fix and it works.

Viicos:
Thanks, handled in https://github.com/python-arq/arq/pull/522.

